PR #36, opened 2018-10-25
UIOR-20 removed api calls since they've been removed on back-end

aliaksei-chumakou:


package.json
```diff
@@ -27,18 +27,14 @@
       "eresource": "1.0",
       "license": "1.0",
       "location": "1.0",
-      "order_format": "1.0",
-      "order_type": "1.0",
       "payment_status": "1.0",
```
craigmcnally: payment_status is gone too, but the mod-orders-storage module descriptor incorrectly still has it listed.  @komick can you please fix the mod-orders-storage MD?

package.json
```diff
@@ -178,13 +178,9 @@
...
           "license.all",
           "location.all",
-          "order_format.all",
-          "order_type.all",
           "physical.all",
-          "receipt_status.all",
           "renewal.all",
-          "source.all",
-          "workflow_status.all"
+          "source.all"
```
craigmcnally: reporting_code.all?

aliaksei-chumakou: @craigmcnally do you mean we need to add this line? I believe it could be done with actual usage of this later

craigmcnally: I wasn't sure.  If we don't need it now adding it later is fine

PR #38, opened 2018-10-26
UIOR-7 remove hierarchy to show list of composite_purchase_orders

aliaksei-chumakou:
This was added before for new mod-orders API /orders, and now this API is changed, so kind of reverting. 
Also, I noticed that comments no longer exist in purchase_order, I believe it's `notes` now, so displaying them in the list


aliaksei-chumakou:
I'm merging this since it's kind a revert and not a big change. Though it's available for post-review

PR #43, opened 2018-11-01
UIOR-37 remove dependency on acquisition_method

aliaksei-chumakou:


aliaksei-chumakou:
@craigmcnally it doesn't look like a separate bug to me, so I've opened a PR https://github.com/folio-org/ui-orders/pull/46

PR #44, opened 2018-11-01
UIOR-28 fix receipt status dropdown values

aliaksei-chumakou:


aliaksei-chumakou:
merging this since a simple fix

PR #49, opened 2018-11-05
UIOR-31 fix PO line details screen open

aliaksei-chumakou:
Now clicking on PO Line item in list of Order's PO Lines will load PO order once more to get a fresh PO line from back-end.


aliaksei-chumakou:
I'm merging this to unblock Andrei's work on PO Line details.
Will for comments to fix them later if any

PR #55, opened 2018-11-09
UIOR-40-Add-Fund-Distribution-accordion

AndreiShumski2:
View
screenshot_7


src/components/POLine/FundDistribution/FundDistributionView.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import _ from 'lodash';
```
aliaksei-chumakou: try to import only needed modules, like `import toString from 'lodash/toString';`

PR #62, opened 2018-11-16
UIOR-38 added I18N and some missing fields

aliaksei-chumakou:
polinedetails
polineedit


src/components/POLine/POLineDetails/POLineDetails.js
```diff
@@ -1,94 +1,138 @@
...
-    const { initialValues } = this.props;
-    return (
+    const { initialValues, intl } = this.props;
+
+    return ([
       <Row>
         <Col xs={6}>
-          <KeyValue label="PO Line ID" value={_.get(initialValues, 'po_line_number')} />
+          <KeyValue
+            label={intl.formatMessage({ id: 'ui-orders.poLine.poLineNumber' })}
```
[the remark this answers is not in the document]
aliaksei-chumakou: @anastasiaHahanova done

src/components/POLine/POLineDetails/POLineDetailsForm.js
```diff
@@ -1,72 +1,273 @@
...
+    const paymentStatuses = [
+      selectPlaceholder,
+      { value: 'Awaiting Payment', label: intl.formatMessage({ id: 'ui-orders.payment_status.awaitingPayment' }) },
+      { value: 'Cancelled', label: intl.formatMessage({ id: 'ui-orders.payment_status.cancelled' }) },
+      { value: 'Fully Paid', label: intl.formatMessage({ id: 'ui-orders.payment_status.fullyPaid' }) },
+      { value: 'Partially Paid', label: intl.formatMessage({ id: 'ui-orders.payment_status.partiallyPaid' }) },
+      { value: 'Payment Not Required', label: intl.formatMessage({ id: 'ui-orders.payment_status.paymentNotRequired' }) },
+      { value: 'Pending', label: intl.formatMessage({ id: 'ui-orders.payment_status.pending' }) },
+    ];
+
```
anastasiaHahanova: I guess, it would be better to place these constants out into the separate file.
Instead of label with formatMessage() can be used only label-id. The file with constants should be imported into current file. `<Field>` Components should be refactored into new syntax with using `<FormattedMessage>`, so dataOptions will be removed: 
https://github.com/folio-org/ui-eholdings/blob/48bf682afc18f43009c97f43b076f49ebf715700/src/components/package/_fields/content-type/package-content-type-field.js

aliaksei-chumakou: @anastasiaHahanova done, please review again

anastasiaHahanova: That's great, that you removed all formatMessage(). Thanks!
However, it would be better to not repeat `<FormattedMessage>` in `<Field>`. Better use `.map()`.

aliaksei-chumakou: @anastasiaHahanova done, please check

PR #63, opened 2018-11-18
Replace stripes.intl.formatMessage with FormattedMessage

cherewaty:
## Purpose
https://github.com/folio-org/stripes-core/pull/489

Retrieving the `intl` object from `this.context.intl` or the `stripes` god object (`stripes.intl`) is a deprecated pattern.

## Approach
Use `<FormattedMessage>`, `<FormattedDate>`, and `<FormattedTime>` as often as possible, only falling back to `formatMessage()`, `formatDate()`, and `formatTime()` (accompanied by `injectIntl()`) when a bare string is absolutely necessary.

### DIFF
--- a/src/settings/index.js
+++ b/src/settings/index.js
@@ -1,29 +1,19 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
 import { Settings } from '@folio/stripes/smart-components';
 import GeneralSettings from './general-settings';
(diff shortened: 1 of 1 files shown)


cherewaty:
Related to https://github.com/folio-org/ui-orders/pull/62

cherewaty:
I don't reviewer-requesting (or merging) permissions on this repo, but @aliaksei-chumakou this looks relevant to your recent work.

PR #64, opened 2018-11-19
UIOR-45-Add-Physical-Resource-Details

AndreiShumski2:
View
screenshot_12
Edit
screenshot_11


src/components/POLine/const.js
```diff
@@ -1,4 +1,5 @@
 export const ERESOURCE = 'Electronic Resource';
+export const PHYSICAL = 'Physical Resource';
 export const PE_MIX = 'P/E Mix';
 
-export const ERESOURCES = [ERESOURCE, PE_MIX];
+export const RESOURCES = [ERESOURCE, PHYSICAL, PE_MIX];
```
aliaksei-chumakou: I don't see the point of this change, `RESOURCES` is never used

src/components/POLine/POLineForm.js
```diff
@@ -166,7 +169,9 @@ class POLineForm extends Component {
       );
     }
 
-    const showEresourcesForm = ERESOURCES.includes(initialValues.order_format);
+    const orderFormat = get(initialValues, 'order_format');
+    const showEresources = (orderFormat === ERESOURCE) || (orderFormat === PE_MIX);
+    const showPhresources = (orderFormat === PHYSICAL) || (orderFormat === PE_MIX);
 
     return (
       <Pane id="pane-poLineForm" defaultWidth="fill" paneTitle={paneTitle} lastMenu={lastMenu} onClose={onCancel} dismissible>
```
anastasiaHahanova: could you, please, place props on its own line when there are more than one

src/components/POLine/POLine.js
```diff
@@ -161,9 +163,11 @@ class POLine extends React.Component {
               <EresourcesView initialValues={initialValues} {...this.props} />
             </Accordion>
           )}
-          <Accordion label="Physical Record Details" id="Physical">
-            <PhysicalView initialValues={initialValues} {...this.props} />
-          </Accordion>
+          {showPhresources && (
```
anton-plotnikov: Do we need extra parentheses here and further in similar cases?

PR #67, opened 2018-11-22
UIOR-41 Updated validation, added FormattedMessage

AndreiShumski2:


src/components/Utils/Validate.js
```diff
@@ -4,8 +4,10 @@ const Required = (value) => {
   return 'Required!';
 };
 
-const RequiredInput = () => {
-  return false;
+const RequiredRefNumber = (value, initialValues) => {
```
aliaksei-chumakou: `initialValues` is a bad name, it doesn't mirror what's in there. Per [docs](https://redux-form.com/7.4.2/examples/fieldlevelvalidation/) it contains `allValues`

src/components/Utils/Validate.js
```diff
@@ -4,8 +4,10 @@ const Required = (value) => {
...
+  const refNumber = initialValues.vendor_detail.ref_number;
+  if (refNumber && !value) return 'Required!';
```
aliaksei-chumakou: make `'Required!'` a constant, it could be changed in future, so we don't need to change it over all usages.

src/components/Utils/Validate.js
```diff
@@ -4,8 +4,10 @@ const Required = (value) => {
...
+  return undefined;
```
aliaksei-chumakou: I think it's more readable to use ternary here:
```
return refNumber && !value
  ? REQUIRED
  : undefined;
```

src/components/Utils/Validate.js
aliaksei-chumakou: `RequiredRefNumber` quite hard to understand, maybe it's better to add comment like
`// field is required only if `vendor_detail.ref_number` isn't empty

src/components/POLine/Vendor/FieldRefNumberType.js
```diff
@@ -0,0 +1,42 @@
...
+
+const REF_NUMBER_TYPE = {
+  suppliersContinuation: "Supplier's continuation order",
+  suppliersUnique: "Supplier's unique order line reference number",
+  agentsUnique: "Agent's unique subscription reference number",
+  librarysContinuation: "Library's continuation order number",
+  internalNumber: 'Internal vendor number',
+};
+
+class FieldRefNumberType extends Component {
```
aliaksei-chumakou: could you please try to use functional component here? it doesn't look like it's going to have an internal state here

src/components/POLine/Vendor/FieldRefNumberType.js
```diff
@@ -0,0 +1,38 @@
...
+      <FormattedMessage
+        id={`ui-orders.vendor.refNumberType.${key}`}
+        key={key}
+      >
+        {(message) => <option value={REF_NUMBER_TYPE[key]}>{message}</option>}
+      </FormattedMessage>
+    ))}
+  </Field>
+);
+
```
aliaksei-chumakou: please add `displayName` as `FieldRefNumberType.displayName = 'FieldRefNumberType';`

src/components/POLine/Vendor/VendorView.js
```diff
@@ -1,30 +1,50 @@
-import React from 'react';
+import React, { Component } from 'react';
+import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
-import _ from 'lodash';
-import { KeyValue, Row, Col } from '@folio/stripes/components';
+import get from 'lodash/get';
```
aliaksei-chumakou: `import { get } from 'lodash';`

src/components/POLine/Vendor/VendorView.js
```diff
@@ -1,30 +1,50 @@
...
+import {
+  Col,
+  KeyValue,
+  Row,
+} from '@folio/stripes/components';
 
-class VendorView extends React.Component {
+class VendorView extends Component {
   static propTypes = {
     initialValues: PropTypes.object
```
aliaksei-chumakou: `initialValues` should be something more descriptive

src/components/POLine/POLine.js
```diff
@@ -164,7 +164,7 @@ class POLine extends Component {
             <br />
           </Accordion>
           <Accordion label="Vendor" id="Vendor">
-            <VendorView initialValues={initialValues} {...this.props} />
+            <VendorView vendor={initialValues.vendor_detail} {...this.props} />
```
aliaksei-chumakou: basically, it's not a `vendor` itself, it's `vendorDetails`... Maybe it's better to change

PR #68, opened 2018-11-22
UIOR-47 added mocks on UI in case they aren't returned from BE

aliaksei-chumakou:


aliaksei-chumakou:
@craigmcnally I don't think it can cause issues as for mocked data. But thanks for looking into that!

PR #87, opened 2018-11-28
UIOR-54 Fund Distribution update Fund Id view

AndreiShumski2:


src/components/POLine/FundDistribution/FundDistributionView.js
```diff
@@ -8,17 +8,20 @@ import {
...
 } from '@folio/stripes/components';
+import FundId from './FundId';
 
 class FundDistributionView extends Component {
   static propTypes = {
     initialValues: PropTypes.shape({
       fund_distribution: PropTypes.arrayOf(PropTypes.object)
+    }),
+    parentResources: PropTypes.shape({
+      fund: PropTypes.arrayOf(PropTypes.object)
```
aliaksei-chumakou: are you sure it's an array here? I think it might be an object of stripes-connect resource

PR #91, opened 2018-12-03
UIOR-11 Settings - Limit number of Purchase Order Lines per PO

AndreiShumski2:
screenshot_2


src/Settings/POLinesLimit.js
```diff
@@ -0,0 +1,139 @@
...
+        replace: PropTypes.func,
+      }),
+    }).isRequired,
+    resources: PropTypes.shape({
+      setPOLinesLimit: PropTypes.shape({
+        records: PropTypes.arrayOf(PropTypes.object),
+      }),
+    }),
+    stripes: PropTypes.shape({
+      intl: PropTypes.object.isRequired,
```
aliaksei-chumakou: it's anti-pattern, please check out docs here https://github.com/folio-org/stripes/blob/master/doc/i18n.md

src/settings/index.js
```diff
@@ -1,26 +1,32 @@
...
+      <Settings
+        {...this.props}
+        pages={this.pages}
+        paneTitle={<FormattedMessage id="ui-orders.settings.index.paneTitle" />}
+      />
     );
   }
 }
+
+export default Orders;
```
aliaksei-chumakou: `Orders` doesn't reflect the meaning of this component. It doesn't display orders or something. Maybe there is a better name. Actually, what's wrong with previous `OrdersSettings`?

src/Settings/POLinesLimit.js
```diff
@@ -0,0 +1,136 @@
...
+      type: 'okapi',
+      records: 'configs',
+      path: 'configurations/entries?query=(module=ORDERS and configName=poLines-limit)',
+      POST: {
+        path: 'configurations/entries',
+      },
+      PUT: {
+        path: 'configurations/entries/%{recordId}',
+      },
+      throwErrors: false,
```
aliaksei-chumakou: I'm not sure if this should be here

PR #101, opened 2018-12-10
UIOR-64 Cost: Clear all quantities when user selects new "PoLine Format"

AndreiShumski2:


src/components/POLine/POLineDetails/POLineDetailsForm.js
```diff
@@ -62,7 +62,7 @@ class POLineDetailsForm extends Component {
             />
           </Col>
           <Col xs={6}>
-            <FieldOrderFormat />
+            <FieldOrderFormat {...this.props} />
```
aliaksei-chumakou: It's better to specify props separately, so we know what is actually used inside. You even specified proptypes only for 2 props

PR #104, opened 2018-12-12
UIOR-46 Close Order

AndreiShumski2:


src/settings/POLinesLimit.js
```diff
@@ -117,7 +120,7 @@ class POLinesLimit extends Component {
                   component={TextField}
                   label={<FormattedMessage id="ui-orders.settings.setPOLInesLimit" />}
                   name="polines_limit"
-                  placeholder="999"
+                  placeholder={get(poLineLimit, 'value', '999')}
```
aliaksei-chumakou: As I can see this code isn't related to the ticket, maybe it make sense to push it within a separate story/ticket. 
But even in that case the logic doesn't look good to me - if value is entered it's better to see it as an input value. not a placeholder

src/components/PurchaseOrder/CloseOrder/CloseOrderModal.js
```diff
@@ -32,18 +34,21 @@ class CloseOrderModal extends Component {
   }
 
   render() {
-    const { orderId } = this.props;
+    const { orderId, workflowStatus } = this.props;
+    const isVisible = (workflowStatus === WORKFLOW_STATUS.open);
```
aliaksei-chumakou:
```suggestion
    const isVisible = orderId && workflowStatus === WORKFLOW_STATUS.open;
```

PR #108, opened 2018-12-14
UIOR-60 display activation due as a date

aliaksei-chumakou:
* use `FormattedDate` to display dates;
* use `format` and `normalize` on Activation Due Field to convert integer to date and vice-versa;
* fill default value from vendor's `expected_activation_interval`;


src/components/LayerCollection/LayerPOLine.js
```diff
@@ -35,6 +36,7 @@ class LayerPOLine extends Component {
     delete newLine.created;
     delete newLine.source;
     delete newLine.vendor_detail;
+    delete newLine.eresource;
```
aliaksei-chumakou: it's temporary, just to not break creation of PO Lines within mod-orders-storage API

PR #109, opened 2018-12-17
UIOR-59 Integrate with mod-orders lines API

aliaksei-chumakou:
* Added mutator and resource to mod-orders Line API;
* works with POST, PUT, DELETE methods;
* fixed missing vendor details instruction field, should be empty as a default value;
* added implementation of Source field, could be used later;
* removed obsolete resources to Lines from mod-orders-storage API;
* removed mock data from UI;


src/components/POLine/POLineDetails/FieldSource.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import { Field } from 'redux-form';
+
+import { Select } from '@folio/stripes/components';
+
+function FieldSource({ sources = [] }) {
```
AndreiShumski2: maybe, it's better to use arrow function for stateless components

aliaksei-chumakou: In this case there is no difference, lets wait common eslint config

PR #110, opened 2018-12-18
UIOR-70 PO: identify a reason for closure

AndreiShumski2:
screenshot_2
screenshot_3


src/settings/ClosingReasons.js
```diff
@@ -0,0 +1,58 @@
...
+import ClosingReasonsForm from './ClosingReasonsForm';
+
+
+class ClosingReasons extends Component {
+  static manifest = Object.freeze({
+    recordId: {},
+    closingReasons: {
+      type: 'okapi',
+      records: 'configs',
+      path: 'configurations/entries?query=(module=ORDERS and configName=closing-reasons)',
```
aliaksei-chumakou: this is specific path for GET, please break it on path and params for GET, no need to use it herer and override for other methods to the default one

src/components/PurchaseOrder/PO.js
```diff
@@ -124,6 +131,26 @@ class PO extends Component {
...
+    const { mutator, resources, onCancel } = this.props;
+    const initialValues = get(resources, ['order', 'records', 0]);
+    const deep = cloneDeep(initialValues);
+
+    delete deep.created_by_name;
+    delete deep.assigned_to_user;
+    delete deep.vendor_name;
+    delete deep.po_lines;
+    mutator.order.PUT({
+      ...deep,
```
aliaksei-chumakou: if you make a deep clone, no need to create new object here - you already have one, `deep`. Also, `deep` - is a bad name, doesn't reflect what contains a variable

src/components/PurchaseOrder/PO.js
```diff
@@ -124,6 +131,26 @@ class PO extends Component {
     return false;
   }
 
+  closeOrder = (values) => {
```
aliaksei-chumakou: `values` is a bad name, doesn't reflect what's the values. This is callback, please be more specific. Or maybe it's better to use Destructuring assignment here: `{reason, note}`

src/components/PurchaseOrder/PO.js
```diff
@@ -124,6 +131,26 @@ class PO extends Component {
...
+    const { mutator, resources, onCancel } = this.props;
+    const initialValues = get(resources, ['order', 'records', 0]);
```
aliaksei-chumakou: `initialValues` is a bad name. It's the prop name for redux form. You don't use it here, it's just the order

src/components/PurchaseOrder/CloseOrder/CloseOrderModal.js
```diff
@@ -51,11 +63,42 @@ class CloseOrderModal extends Component {
...
+              />
+            </Col>
+          </Row>
+          <Button
+            buttonStyle="primary"
+            onClick={() => {
+              closeOrderSubmit(this.state.values);
+              this.closeModal();
+            }}
+            style={{ float: 'right' }}
```
aliaksei-chumakou: I don't think using float is a common practice in Folio. I'm pretty sure there is a possibility to reach the same goal in other way

src/components/PurchaseOrder/CloseOrder/CloseOrderModal.js
```diff
@@ -51,11 +63,42 @@ class CloseOrderModal extends Component {
...
-        />
+        >
+          <Row>
+            <Col xs={12}>
+              <Select
+                placeHolder="---Choose the reason---"
+                dataOptions={reasons}
+                required
+                label={<FormattedMessage id="ui-orders.closeOrderModal.reason" />}
+                onChange={(e => this.setState({ values: { reason: e.target.value } }))}
```
aliaksei-chumakou: please describe the state default values in constructor. And use specific attributes for reason and note, no need to use nesting and object like `values`.

src/components/PurchaseOrder/CloseOrder/CloseOrderModal.js
```diff
@@ -51,11 +63,42 @@ class CloseOrderModal extends Component {
...
+              closeOrderSubmit(this.state.values);
+              this.closeModal();
+            }}
+            style={{ float: 'right' }}
+          >
+            <FormattedMessage id="ui-orders.closeOrderModal.submit" />
+          </Button>
+          <Button
+            onClick={() => this.closeModal()}
+            style={{ float: 'right', marginRight: '10px' }}
```
aliaksei-chumakou: again, are you sure that layout by floats is a common way in Folio? I don't think so...

src/components/LayerCollection/LayerPO.js
```diff
@@ -25,16 +27,11 @@ class LayerPO extends Component {
...
+    const { parentMutator, onCancel } = this.props;
 
-    delete deep.created_by_name;
-    delete deep.assigned_to_user;
-    delete deep.vendor_name;
-    delete deep.po_lines;
-    this.props.parentMutator.records.PUT(deep).then(() => {
-      this.props.onCancel();
-    });
+    updateOrderResource(order, parentMutator.records)
```
aliaksei-chumakou: try to make this function `async` and `await updateOrderResource(order, parentMutator.records)`

src/components/Utils/updateOrderResource.js
```diff
@@ -0,0 +1,16 @@
...
+
+const updateOrderResource = (order, mutator, changedProps) => {
+  const cloneOrder = cloneDeep(order);
+
+  delete cloneOrder.created_by_name;
+  delete cloneOrder.assigned_to_user;
+  delete cloneOrder.vendor_name;
+  delete cloneOrder.po_lines;
+
+  const newOrder = Object.assign(cloneOrder, changedProps);
```
aliaksei-chumakou: basically, yo don't need `newOrder`. just `Object.assign(cloneOrder, changedProps);`

src/components/Utils/updateOrderResource.js
```diff
@@ -0,0 +1,16 @@
+import { cloneDeep } from 'lodash';
+
+const updateOrderResource = (order, mutator, changedProps) => {
+  const cloneOrder = cloneDeep(order);
```
aliaksei-chumakou: `clonedOrder` might be a better name

src/components/PurchaseOrder/PO.js
```diff
@@ -132,6 +152,8 @@ class PO extends Component {
       <PaneMenu>
         <IfPermission perm="purchase_order.item.put">
           <CloseOrderModal
+            closeOrderSubmit={this.closeOrder}
+            closingReasons={parentResources.closingReasons.records}
```
aliaksei-chumakou: I don't think you need to use `parentResources` here. It could be done through `resources`

src/settings/ClosingReasons.js
```diff
@@ -0,0 +1,63 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import ClosingReasonsForm from './ClosingReasonsForm';
+
+
+class ClosingReasons extends Component {
+  static manifest = Object.freeze({
+    recordId: {},
```
aliaksei-chumakou: you don't need `recordId` at all, you can reference lines limit setting as an example.

src/settings/ClosingReasons.js
```diff
@@ -0,0 +1,63 @@
...
+      type: 'okapi',
+      records: 'configs',
+      path: 'configurations/entries',
+      GET: {
+        params: {
+          query: '(module=ORDERS and configName=closing-reasons)',
+        },
+      },
+      POST: {
+        path: 'configurations/entries',
```
aliaksei-chumakou: this makes no sense, it's the same as common `path` for the resource

src/settings/ClosingReasons.js
```diff
@@ -0,0 +1,63 @@
...
+      },
+      PUT: {
+        path: 'configurations/entries/%{recordId}',
```
aliaksei-chumakou: see comment above - you don't need `recordId`

src/settings/ClosingReasons.js
```diff
@@ -0,0 +1,63 @@
...
+      },
+      DELETE: {
+        path: 'configurations/entries/%{recordId}',
```
aliaksei-chumakou: the same as above

src/settings/ClosingReasons.js
```diff
@@ -0,0 +1,63 @@
...
+    };
+  }
+
+  render() {
+    const { label, resources } = this.props;
+
+    return (
+      <div style={this.styles.closingReasonsWrapper}>
+        <ClosingReasonsForm
+          initialValues={resources.closingReasons}
```
aliaksei-chumakou: I think you should extract this to variable, like `const initialValues = get(resources, 'closingReasons', []);`

src/settings/ClosingReasons.js
```diff
@@ -0,0 +1,63 @@
...
+          title={label}
+          {...this.props}
```
aliaksei-chumakou: this is anti-pattern

src/settings/ClosingReasonsForm.js
```diff
@@ -0,0 +1,169 @@
...
+            onClick={() => this.onRemoveClosingReason(fields, index)}
+          >
+            <FormattedMessage id="ui-orders.settings.closingReasons.removeBtn" />
+          </Button>
+        </Col>
+      </Row>
+    );
+  }
+
+  onSaveClosingReason = (index) => {
```
aliaksei-chumakou: this should be above in the component hierarchy. Basically, form is Presentational component. Just pass callback to it via props

src/settings/ClosingReasonsForm.js
```diff
@@ -0,0 +1,169 @@
...
+    const { stripes: { store }, resources } = this.props;
+    const formValues = getFormValues('closing-reasons')(store.getState());
```
aliaksei-chumakou: you don't need this at all. I think you need to re-think your approach with sub-forms, they have submit button/callback, and they pass needed values to it. You can reference redux-form example https://redux-form.com/7.4.2/examples/fieldarrays/

PR #111, opened 2018-12-18
UIOR-79 Setting for PO Line Limit not actually being saved

AndreiShumski2:


aliaksei-chumakou:
I think it could be fixed with refactoring to make things easy https://github.com/folio-org/ui-orders/pull/112/files

PR #118, opened 2018-12-22
UIOR-75 Add POLine limit exceeded error message

AndreiShumski2:


src/components/LayerCollection/LayerPOLine.js
```diff
@@ -30,16 +32,56 @@ class LayerPOLine extends Component {
...
+  }
+
+  submitPOLine = async (data) => {
     const newLine = cloneDeep(data);
     const { lineMutator, onCancel } = this.props;
 
-    lineMutator.POST(newLine).then(() => {
+    try {
+      const response = await lineMutator.POST(newLine);
+      if (!response.id) throw new Error(response);
```
aliaksei-chumakou: Are you sure that object structure will be the same? I don;t think so. Please describe the purpose of this line

src/components/LayerCollection/LayerPOLine.js
```diff
@@ -30,16 +32,56 @@ class LayerPOLine extends Component {
...
       onCancel();
+    } catch (e) {
+      const errors = await e.json();
+      if (errors.errors.find(el => el.code === 'lines_limit')) this.openModal();
```
aliaksei-chumakou: please don't write like, it's unreadable. And here could be an error - `errors` response structure is not guaranteed.

src/components/LayerCollection/LayerPOLine.js
```diff
@@ -30,16 +32,56 @@ class LayerPOLine extends Component {
...
+    }
+  }
+
+  createOrder = async () => {
+    const { order, parentMutator } = this.props;
+    const newOrder = cloneDeep(order);
+
+    newOrder.created = moment.utc().format();
```
aliaksei-chumakou: please remove this line, it makes no sense. This field is generated on back-end

src/components/LayerCollection/LayerPOLine.js
```diff
@@ -30,16 +32,56 @@ class LayerPOLine extends Component {
...
+    delete newOrder.adjustment;
+    delete newOrder.assigned_to_user;
+    delete newOrder.created_by_name;
```
aliaksei-chumakou: copy-paste I feel, please use util function for common code.It already has mutator, so it's enough to check if object has id - use PUT, if no - use POST

src/components/LayerCollection/LayerPOLine.js
```diff
@@ -57,32 +57,30 @@ class LayerPOLine extends Component {
     const { lineMutator, onCancel } = this.props;
 
     try {
-      const response = await lineMutator.POST(newLine);
-      if (!response.id) throw new Error(response);
+      await lineMutator.POST(newLine);
       onCancel();
     } catch (e) {
       const errors = await e.json();
```
aliaksei-chumakou: instead of `errors` you get here JSON representation of response. So `response` might be a better name.
https://developer.mozilla.org/en-US/docs/Web/API/Body/json

src/components/LayerCollection/LayerPOLine.js
```diff
@@ -57,32 +57,30 @@ class LayerPOLine extends Component {
...
-      if (errors.errors.find(el => el.code === 'lines_limit')) this.openModal();
+      if (errors.errors.find(el => el.code === 'lines_limit')) {
```
aliaksei-chumakou: still, first `errors` could be empty object.

src/components/LayerCollection/LayerPOLine.js
```diff
@@ -57,32 +57,30 @@ class LayerPOLine extends Component {
...
+        this.openModal();
+      } else {
```
aliaksei-chumakou: I don't see too much sense in this `else`

src/components/LayerCollection/LayerPOLine.js
```diff
@@ -57,32 +57,30 @@ class LayerPOLine extends Component {
...
-    });
+
+    try {
+      const newOrder = await createOrder(order, parentMutator.records);
+      parentMutator.query.update({
+        _path: `/orders/view/${newOrder.id}`,
+        layer: null,
+      });
+    } catch (e) {
+      console.error(e);
```
aliaksei-chumakou: I think we might need to find a way to notify user that order was not created, console.error looks like mock

src/routes/Main.js
```diff
@@ -202,23 +191,18 @@ class Main extends Component {
...
-    mutator.records.POST(order).then(newOrder => {
+    try {
+      const newOrder = await createOrder(order, mutator.records);
       mutator.query.update({
         _path: `/orders/view/${newOrder.id}`,
         layer: null,
       });
-    });
+    } catch (e) {
+      console.error(e);
```
aliaksei-chumakou: the same as above, it's better to show user something

src/routes/Main.js
```diff
@@ -192,23 +191,18 @@ class Main extends Component {
...
-    order.created = moment.utc().format();
-    delete order.created_by_name;
-    delete order.assigned_to_user;
-    delete order.vendor_name;
-    delete order.bill_to;
-    delete order.ship_to;
-
-    mutator.records.POST(order).then(newOrder => {
+    try {
+      const newOrder = await createOrder(order, mutator.records);
```
aliaksei-chumakou: `mutator.records` ? are you sure?

src/components/Utils/createOrder.js
```diff
@@ -0,0 +1,23 @@
+import { cloneDeep } from 'lodash';
+
+const createOrder = (order, mutator) => {
+  const newOrder = cloneDeep(order);
+
+  delete newOrder.assigned_to_user;
+  delete newOrder.created_by_name;
+  delete newOrder.vendor_name;
+  if (newOrder.id) {
```
aliaksei-chumakou: I don't think it's needed here, what I meant - to check this in some common `saveOrder` function to choose appropriate method of mutator - PUT or POST. If you have here only the POST, so no need to this check

src/routes/Main.js
AndreiShumski2: yes, I am

src/components/Utils/orderResource.js
```diff
@@ -0,0 +1,38 @@
...
+
+  Object.assign(clonedOrder, changedProps);
+
+  return saveOrder(clonedOrder, mutator);
+};
+
+export const createOrderResource = (order, mutator) => {
+  const clonedOrder = cloneDeep(order);
+
+  delete clonedOrder.bill_to;
```
aliaksei-chumakou: I think this should be inside `saveOrder()`

src/components/LayerCollection/LayerPOLine.js
```diff
@@ -30,17 +36,57 @@ class LayerPOLine extends Component {
...
+  openLineLimitExceededModal = (line) => {
+    this.setState({
+      openModal: true,
+      line,
+    });
+  }
+
+  closeLineLimitExceededModal = () => {
+    this.setState({
+      openModal: false,
```
aliaksei-chumakou: shouldn't you clear `line` state prop here?

src/components/LayerCollection/LayerPOLine.js
```diff
@@ -30,17 +36,57 @@ class LayerPOLine extends Component {
 
   constructor(props) {
     super(props);
+
+    this.state = {
+      openModal: false,
```
aliaksei-chumakou: I think it's a good practice to specify default values for all state props

PR #129, opened 2019-01-02
UIOR-66 POLine Type new Title or select a Title from inventory instances

AndreiShumski2:
screenshot_5

Overview: When creating an order the user may be ordering additional copies of a title they already own and have represented in inventory.

High Level Requirements:

Add new title by typing in the "Title" field
Ability to select an existing title from inventory by click the plus and using the search modal to select a title

### DIFF
--- a/src/components/POLine/Item/ItemForm.js
+++ b/src/components/POLine/Item/ItemForm.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
 import { Field } from 'redux-form';
+
 import {
   Col,
   Datepicker,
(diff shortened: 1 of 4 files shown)


src/components/POLine/Item/ItemForm.js
```diff
@@ -1,128 +1,173 @@
...
+      materialTypes: PropTypes.object.isRequired,
+    }).isRequired,
+    dispatch: PropTypes.func.isRequired,
+    change: PropTypes.func.isRequired,
+  }
+
+  onAddInstance = (instance) => {
+    const { dispatch, change } = this.props;
+
+    dispatch(change('title', `${instance.title}`));
```
aliaksei-chumakou: no need for string templates, just instance.title

src/components/POLine/Item/ItemForm.js
```diff
@@ -1,128 +1,173 @@
...
+              fullWidth
+              label={<FormattedMessage id="ui-orders.itemDetails.title" />}
+              name="title"
+            />
+          </Col>
+          <Col xs={1}>
+            {this.selectInstanceModal()}
+          </Col>
+        </Row>
+        <Row>
```
aliaksei-chumakou: it doesn't make too much sense to me using `Row` like this. It will be the same if you just add `<Col xs={12}>`

AndreiShumski2: updated
screenshot_6

PR #134, opened 2019-01-04
UIOR-89 Order Settings: Purchase Order Number

AndreiShumski2:
screenshot_7
Purpose: 
-create a new setting for Orders - Purchase Order Number.
-create a checkbox for the ability to Edit Purchase Order Number when creating order (if false users must use the auto-generated PO Number)


src/settings/OrderNumber.js
```diff
@@ -0,0 +1,49 @@
...
+  };
+
+  constructor(props) {
+    super(props);
+    this.configManager = props.stripes.connect(ConfigManager);
+  }
+
+  getInitialValues = (settings) => {
+    const value = settings.length === 0 ? '' : settings[0].value;
+    const defaultConfig = { isEditPONumber: false };
```
aliaksei-chumakou: `canUserEditOrderNumber` sounds more clear, as for me.

src/settings/OrderNumber.js
```diff
@@ -0,0 +1,49 @@
...
+    return config;
+  }
+
+  render() {
+    const { label, stripes } = this.props;
+
+    return (
+      <this.configManager
+        label={label}
+        moduleName="ORDERS"
```
aliaksei-chumakou: We have to make this a constant

src/settings/OrderNumber.js
```diff
@@ -0,0 +1,49 @@
...
+        configName="poNumber"
```
aliaksei-chumakou: if there is no requirements on this, I think `order.number` is a better configuration name

src/settings/OrderNumberForm.js
```diff
@@ -0,0 +1,98 @@
...
+  Pane,
+  Row,
+} from '@folio/stripes/components';
+
+class OrderNumberForm extends Component {
+  static propTypes = {
+    stripes: PropTypes.shape({
+      store: PropTypes.object.isRequired
+    }),
+    handleSubmit: PropTypes.func.isRequired,
```
aliaksei-chumakou: actually since you're not overriding functionality, you can pass you controls as a children, without wrapping them into a `<form>`. What do you think?

PR #139, opened 2019-01-08
UIOR-59 fix saving Line

aliaksei-chumakou:


src/index.js
```diff
@@ -43,4 +42,4 @@ class Orders extends Component {
   }
 }
 
-export default hot(module)(Orders);
```
aliaksei-chumakou: since hot relodaing isn't working, it's safe to remove for faster builds

PR #146, opened 2019-01-10
UIOR-96 Order Settings: Purchase Order Number - Prefix & Suffix

AndreiShumski2:
screenshot_8
Purpose: The user must have the ability to adjust what prefixes and suffixes can be used when building orders. In settings, administrative users will be able to add prefixes and suffixes to the respective lists that will appear in a select list for other users when creating orders.

-create accordion for the ability to select Prefix & Suffix for Purchase Order Number when creating order

Story https://issues.folio.org/browse/UIOR-77
Sub-task https://issues.folio.org/browse/UIOR-96


src/settings/OrderNumberForm.js
```diff
@@ -0,0 +1,159 @@
...
+        disabled={(pristine || submitting)}
+        marginBottom0
+      >
+        <FormattedMessage id="ui-orders.settings.poNumber.saveBtn" />
+      </Button>
+    );
+  };
+
+  addPrefix = ({ inputValue }) => {
+    // eslint-disable-next-line react/no-access-state-in-setstate
```
aliaksei-chumakou: Actually, as a developer you're not supposed to silent lint rules, it's their purpose to check such situations for a reason. 
In this very case I think there is no reason to make the silent and it's OK to use `setState((oldSate) => newState)`

src/settings/OrderNumberForm.js
```diff
@@ -0,0 +1,159 @@
...
+    this.setState({ prefix: [...this.state.prefix, { label: inputValue, value: inputValue }] });
+  }
+
+  addSuffix = ({ inputValue }) => {
+    // eslint-disable-next-line react/no-access-state-in-setstate
```
aliaksei-chumakou: the same as above

src/settings/OrderNumber.js
```diff
@@ -40,26 +47,9 @@ class OrderNumber extends Component {
...
-                label={<FormattedMessage id="ui-orders.settings.poNumber.editPONumber" />}
-                name="orderNumber"
-                type="checkbox"
-              />
-            </Col>
-          </Row>
-        </Accordion>
-      </this.configManager>
+        onBeforeSave={this.beforeSave}
+        configFormComponent={OrderNumberForm}
```
aliaksei-chumakou: Is there a reason to use `configFormComponent` ? What I see - it's just duplicates the code for form and save button. Previous way was less duplicating, IMHO.

src/settings/OrderNumberForm.js
```diff
@@ -0,0 +1,159 @@
...
+    pristine: PropTypes.bool,
+    submitting: PropTypes.bool,
+    initialValues: PropTypes.object,
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      prefix: [],
```
aliaksei-chumakou: What is the reason of having this state? I think it could be managed by redux-form `Field`
```
                    <Field
                      name="tags"
                      component={MultiSelection}
                      emptyMessage="Empty message"
                      fullWidth
                      dataOptions={OPTIONS}
                      actions={actions}
                    />
```

src/settings/OrderNumberForm.js
```diff
@@ -0,0 +1,170 @@
...
+    dispatch(change('prefix', selectedPrefix));
+  }
+
+  addSuffix = ({ inputValue }) => {
+    const { dispatch, change, stripes: { store } } = this.props;
+    const suffixList = this.getOptionsList('suffixList');
+    const formValues = getFormValues('orderNumberForm')(store.getState());
+    const selectedSuffix = get(formValues, 'prefix', []);
+
+    suffixList.push({ label: inputValue, value: inputValue });
```
aliaksei-chumakou: I think you might need to use `...` here to not modify existing array. Because you're dispatching the new value later.
```
const newOption = {
  label: inputValue,
  value: inputValue,
};
const newSuffixes = [newOption, ...suffixList];`

File: src/settings/OrderNumber.js [hunk 6]
```diff
@@ -22,44 +14,61 @@ class OrderNumber extends Component {
 
   constructor(props) {
     super(props);
+
     this.configManager = props.stripes.connect(ConfigManager);
   }
 
   getInitialValues = (settings) => {
-    const value = settings.length && settings[0].value === 'true';
+    const value = settings.length === 0 ? '' : settings[0].value;
```
aliaksei-chumakou: I'm not sure that I understand what is this for `settings[0].value === 'true'`... 
`const orderNumberSetting = get(settings, '0.value', '{}');`

src/settings/OrderNumber.js
```diff
@@ -22,44 +14,61 @@ class OrderNumber extends Component {
...
+    const defaultConfig = {
+      orderNumber: false,
```
aliaksei-chumakou: this name is not clear at all, IMHO. It's a spill-over from previous PR https://github.com/folio-org/ui-orders/pull/134#discussion_r245263010

src/settings/OrderNumber.js
```diff
@@ -22,44 +14,61 @@ class OrderNumber extends Component {
...
+      selectedPrefixes: [],
+      prefixesList: [],
```
aliaksei-chumakou: `prefixes` is already enough to show that it's not an object

src/settings/OrderNumber.js
```diff
@@ -22,44 +14,61 @@ class OrderNumber extends Component {
...
   constructor(props) {
     super(props);
+
     this.configManager = props.stripes.connect(ConfigManager);
   }
 
   getInitialValues = (settings) => {
-    const value = settings.length && settings[0].value === 'true';
+    const value = settings.length === 0 ? '' : settings[0].value;
+    const defaultConfig = {
```
aliaksei-chumakou: if you create it here, just name it after the setting - `config`. You can extend it later with values from back-end

src/settings/OrderNumber.js
```diff
@@ -22,44 +14,61 @@ class OrderNumber extends Component {
...
+      orderNumber: false,
+      selectedPrefixes: [],
+      prefixesList: [],
+      selectedSuffixes: [],
+      suffixesList: [],
+    };
+    let config;
+
+    try {
+      config = Object.assign({}, defaultConfig, JSON.parse(value));
```
aliaksei-chumakou: Actually it's better to leave in try-catch block only code with parsing error expected:
```
orderNumberSetting = JSON.parse(orderNumberSetting)
```
in catch you can assign `{}` to `orderNumberSetting`.
and after just make `Object.assign(config, orderNumberSetting);`

src/settings/OrderNumber.js
```diff
@@ -22,44 +14,61 @@ class OrderNumber extends Component {
...
+      selectedPrefixes,
+      prefixesList,
+      selectedSuffixes,
+      suffixesList,
+    } = data;
 
-    return { orderNumber: value };
+    return JSON.stringify({
+      orderNumber,
+      selectedPrefixes,
```
aliaksei-chumakou: I hope you're sending to back-end a plain lists of strings? like `['OO', 'OO2']`

AndreiShumski2:
@aliaksei-chumakou Thank you for your review and help

PR #153, opened 2019-01-14
UIOR-83 Add PO Line - Vendor Ref Type required field

AndreiShumski2:
Purpose:
Vendor Ref Number and Vendor Ref Type are paired fields.
If Vendor Ref Number (free text field) is filled in, then Vendor Ref Type should be required.
If Vendor Ref Number is not filled in, then Vendor Ref Type should not be required.
screenshot_11
screenshot_10

### DIFF
--- a/src/components/Utils/Validate.js
+++ b/src/components/Utils/Validate.js
@@ -15,6 +15,15 @@ export const requiredRefNumber = (value, allValues) => {
     : undefined;
 };
 
+// Field is required only if 'vendor_detail.ref_number_type' isn't empty
+export const requiredRefNumberType = (value, allValues) => {
+  const refNumberType = allValues.vendor_detail.ref_number_type;
+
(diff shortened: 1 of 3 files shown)


src/components/Utils/Validate.js
```diff
@@ -15,6 +15,15 @@ export const requiredRefNumber = (value, allValues) => {
     : undefined;
 };
 
+// Field is required only if 'vendor_detail.ref_number_type' isn't empty
+export const requiredRefNumberType = (value, allValues) => {
+  const refNumberType = allValues.vendor_detail.ref_number_type;
```
aliaksei-chumakou: Have you checked this on new Line? I think that `vendor_detail` might be absent and this line might cause an exception

AndreiShumski2: Checked. It works with new Line too.

PR #170, opened 2019-01-17
UIOR-68 Add Renewal accordion

aliaksei-chumakou:


src/components/PurchaseOrder/renewals/RenewalsForm.js
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import {
+  Field,
+} from 'redux-form';
```
AndreiShumski2: It's possible to write in one line

PR #181, opened 2019-01-22
UIOR-75 Show PO line limit exceeded error message

AndreiShumski2:
screenshot_17

### DIFF
--- a/src/components/PurchaseOrder/PO.js
+++ b/src/components/PurchaseOrder/PO.js
@@ -8,6 +8,7 @@ import {
   Accordion,
   AccordionSet,
   Button,
+  Callout,
   Col,
   ExpandAllButton,
   Icon,
@@ -22,12 +23,20 @@ import {
   LayerPOLine,
 } from '../LayerCollection';
 import {
+  CONFIG_API,
   LINES_API,
   ORDER_DETAIL_API,
 } from '../Utils/api';
 import transitionToParams from '../Utils/transitionToParams';
-import { updateOrderResource } from '../Utils/orderResource';
-import { MODULE_ORDERS } from '../Utils/const';
+import {
+  cloneOrder,
+  updateOrderResource,
+} from '../Utils/orderResource';
+import {
+  MODULE_ORDERS,
+  CONFIG_LINES_LIMIT,
+  CONFIG_CLOSING_REASONS,
+} from '../Utils/const';
 import { ORDER_TYPE } from './PODetails/FieldOrderType';
 import CloseOrderModal from './CloseOrder';
 import { WORKFLOW_STATUS } from './Summary/FieldWorkflowStatus';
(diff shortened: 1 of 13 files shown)


src/components/PurchaseOrder/PO.js
```diff
@@ -59,6 +64,16 @@ class PO extends Component {
         },
       },
     },
+    linesLimit: {
+      type: 'okapi',
+      records: 'configs',
+      path: 'configurations/entries',
```
aliaksei-chumakou: please use constant for paths, it's expected to be reused across the application

src/components/PurchaseOrder/PO.js
```diff
@@ -59,6 +64,16 @@ class PO extends Component {
...
+      GET: {
+        params: {
+          query: `(module=${MODULE_ORDERS} and configName=poLines-limit)`,
```
aliaksei-chumakou: please use constant for such things as config names, it's also expected to use in several places

src/components/PurchaseOrder/PO.js
```diff
@@ -137,8 +153,16 @@ class PO extends Component {
   }
 
   onAddPOLine = (e) => {
+    const { resources } = this.props;
+    const linesLimit = Number(get(resources, ['linesLimit', 'records', '0', 'value'], '1'));
```
aliaksei-chumakou: Firstly, `'1'` makes not much sense to me.
Secondly - why back-end doesn't return a number, since it's a number-typed config?

src/components/PurchaseOrder/PO.js
```diff
@@ -137,8 +153,16 @@ class PO extends Component {
...
+    const poLines = get(resources, ['order', 'records', '0', 'po_lines'], []);
+
     if (e) e.preventDefault();
```
aliaksei-chumakou: I'm not sure if you need to pass `e` in this function and run this line. The question is - what is trying to be prevented here?

src/components/Utils/orderResource.js
```diff
@@ -31,7 +31,12 @@ export const cloneOrder = (order, mutator, line) => {
 
   delete clonedOrder.id;
   delete clonedOrder.adjustment;
-  clonedOrder.po_lines = [line];
```
aliaksei-chumakou: I still can't see removing of po_number fix

src/components/Utils/orderResource.js
```diff
@@ -31,7 +31,12 @@ export const cloneOrder = (order, mutator, line) => {
...
+  if (line) {
+    clonedOrder.po_lines = [line];
+    delete clonedOrder.po_lines[0].purchase_order_id;
```
aliaksei-chumakou: it looks to me as over-engineering. Just move this line above and make it simpler:
`delete line.purchase_order_id;`

src/components/PurchaseOrder/PO.js
AndreiShumski2: in mod-config schema value is defined as a string
https://s3.amazonaws.com/foliodocs/api/mod-configuration/config.html#configurations_entries_post

PR #185, opened 2019-01-23
UIOR-88 PO: Inconsistent behavior of Vendor Field

AndreiShumski2:
screenshot_1
screenshot_2
screenshot_3
https://issues.folio.org/browse/UIOR-88

### DIFF
--- a/src/components/PurchaseOrder/PODetails/PODetailsForm.js
+++ b/src/components/PurchaseOrder/PODetails/PODetailsForm.js
@@ -18,10 +18,11 @@ import {
 } from '@folio/stripes/components';
 import { Pluggable } from '@folio/stripes/core';
 
-import FieldOrderType from './FieldOrderType';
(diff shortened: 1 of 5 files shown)


src/components/PurchaseOrder/PODetails/PODetailsForm.js
```diff
@@ -271,10 +296,17 @@ class PODetailsForm extends Component {
...
-          <Col xs={6} md={3}>
+          <Col
+            xs={6}
+            lg={3}
+          >
             <FieldOrderType />
           </Col>
-          <Col xs={12}>
+          <Col
+            xs={8}
```
aliaksei-chumakou: don't you think that it's kind of hard to keep in mind all these numbers of `xs`? It's not obvious where field will be on a new line. Maybe it makes sense to break up on lines using `<Row>` tags?

PR #188, opened 2019-01-24
Update translation strings as of 20190124T161902

folio-translations:
Please do not manually merge this pull request.  It will either be automatically merged or withdrawn.


aliaksei-chumakou:
@dltj it would be great to have this merged!

PR #193, opened 2019-01-25
UIREC-7 Create Receive Button

AndreiShumski2:
screenshot_1

https://issues.folio.org/browse/UIREC-7

### DIFF
--- a/src/components/PurchaseOrder/PO.js
+++ b/src/components/PurchaseOrder/PO.js
@@ -9,7 +9,6 @@ import {
   AccordionSet,
   Button,
   Callout,
-  Col,
   ExpandAllButton,
   Icon,
   IconButton,
@@ -47,6 +46,7 @@ import { PODetailsView } from './PODetails';
 import { SummaryView } from './Summary';
 import { RenewalsView } from './renewals';
 import LinesLimit from './LinesLimit';
+import css from './PO.css';
 
 class PO extends Component {
   static manifest = Object.freeze({
@@ -120,7 +120,6 @@ class PO extends Component {
       isCloseOrderModalOpened: false,
       isLinesLimitExceededModalOpened: false,
     };
-    this.onAddPOLine = this.onAddPOLine.bind(this);
     this.transitionToParams = transitionToParams.bind(this);
   }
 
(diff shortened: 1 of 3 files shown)


src/components/PurchaseOrder/PO.js
```diff
@@ -242,10 +241,21 @@ class PO extends Component {
     const orderNumber = get(order, 'po_number', '');
     const poLines = get(order, 'po_lines', []);
     const workflowStatus = get(order, 'workflow_status');
-    const isCloseOrderButtonVisible = orderId && workflowStatus === WORKFLOW_STATUS.open;
+    const isWorkflowStatusOpen = workflowStatus === WORKFLOW_STATUS.open;
+    const isCloseOrderButtonVisible = orderId && isWorkflowStatusOpen;
+    const isReceiveButtonVisible = orderId && isWorkflowStatusOpen && (poLines.length > 0);
```
aliaksei-chumakou: I think `poLines.length` is not enough. On the meeting Dennis mentioned physical quantity and electronic. So it looks like you supposed to look (maybe reduce) `poLines` to have these values more than 0

src/components/PurchaseOrder/PO.js
```diff
@@ -242,10 +241,21 @@ class PO extends Component {
...
     const lastMenu = (
       <PaneMenu>
         <IfPermission perm="purchase_order.item.put">
+          {isReceiveButtonVisible && (
+            <Button
+              buttonStyle="primary"
+              marginBottom0
+              style={{ marginRight: '10px' }}
```
aliaksei-chumakou: these `marginRight` don't work well with RTL feature. Maybe it makes sense to use flebox spacing here between buttons

src/components/PurchaseOrder/PO.js
AndreiShumski2: Ok. Now physical and/or electronic quantity are required fields to fill in any case.

aliaksei-chumakou: The key word here `Now`. Who knows what order types could be in a future, and Dennis was clear regarding to look at quantity fields

AndreiShumski2: I got it, thank you.

PR #200, opened 2019-01-28
UIOR-115 Update the po_number schema to use camelCase

AndreiShumski2:
FOLIO naming conventions for schemas dictate that camelCase is used for field names instead of under_score notation.
https://issues.folio.org/browse/UIOR-115
Should be merged after
https://issues.folio.org/browse/MODORDERS-130
https://github.com/folio-org/acq-models/pull/80
https://github.com/folio-org/mod-orders-storage/pull/73
https://github.com/folio-org/mod-orders/pull/84

### DIFF
--- a/src/components/Utils/orderResource.js
+++ b/src/components/Utils/orderResource.js
@@ -29,10 +29,10 @@ export const updateOrderResource = (order, mutator, changedProps) => {
 
 export const createOrderResource = (order, mutator) => {
   const clonedOrder = cloneDeep(order);
-  const { numberPrefix = '', numberSuffix = '', po_number: orderNumber = '' } = clonedOrder;
-  const fullOrderNumber = `${numberPrefix}${orderNumber}${numberSuffix}`.trim();
+  const { numberPrefix = '', numberSuffix = '', poNumber = '' } = clonedOrder;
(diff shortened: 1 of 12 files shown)


src/components/Utils/orderResource.js
```diff
@@ -29,10 +29,10 @@ export const updateOrderResource = (order, mutator, changedProps) => {
 
 export const createOrderResource = (order, mutator) => {
   const clonedOrder = cloneDeep(order);
-  const { numberPrefix = '', numberSuffix = '', po_number: orderNumber = '' } = clonedOrder;
+  const { numberPrefix = '', numberSuffix = '', poNumber: orderNumber = '' } = clonedOrder;
```
aliaksei-chumakou: in case of `poNumber` there is no need in `orderNumber`, it could be just `poNumber = ''`

translations/ui-orders/en.json
```diff
@@ -399,7 +399,7 @@
   "order.paneTitle.details": "Purchase Order - {orderNumber}",
   "order.paneTitle.detailsLoading": "Details",
   "order.paneTitle.edit": "Edit - {orderNumber}",
-  "order.po_number": "PO Number",
+  "order.poNumber": "PO Number",
```
aliaksei-chumakou: Let's not change label key, due to complex way to get this on UI for en_US

PR #326, opened 2019-03-21
UIOR-183 Allow 0.0 for unit price in Cost for Orders

aliaksei-chumakou:
## Purpose
Allow 0.0 for unit price in Cost for Orders.
https://issues.folio.org/browse/UIOR-183

## Approach
Basically, there is a different validations on UI for quantities and prices, since required prices can be 0 or `0.00` and required quantities should be more or equal 1

## Pre-Merge Checklist


src/components/POLine/Cost/CostForm.js
```diff
@@ -24,16 +23,33 @@ import {
...
 
+// Validation for Fields with type 'number' requires positive integer
+export const requiredPositiveQuantity = (value) => {
+  return value >= 1
+    ? undefined
+    : <FormattedMessage id="ui-orders.cost.validation.shouldBePositive" />;
+};
+
+const validateRequiredNotNegative = (value) => {
+  return value === 0 || value > 0
```
craigmcnally: could this be shortened to simply ` value < 0 ? <formattedMessage ...> : undefined`?  Was it done this way for consistency with the other ternary expressions used in the other validation functions?

aliaksei-chumakou: basically in JS the way you've suggested is doing a quite different comparison. we are not sure if value is a `Number` type. So empty string will pass through as a good value in your case

craigmcnally: Thanks makes sense now... but doesn't that mean that `'0'` would be considered invalid, while `0` or `'2'` are considered valid?

aliaksei-chumakou: That's true, but in overall `return value === 0 || value > 0` precisely describes what we expect as truthy values, so for readability purposes I prefer to leave it as is. 
If you look into `parse={parseMoney}` field attribute, which calls before validate, it return `Number` or some falsy value (empty string included), what we filter here as not valid. Hope this makes sense!

NikitaSedyx: hm, what about `value >= 0`?

NikitaSedyx: doesn't work with empty string, but fine for undefined

NikitaSedyx:
:+1:

src/components/POLine/Cost/CostForm.js
craigmcnally: Right, `value === 0 || value > 0` gets us 99% of the way there, the case I mention is just an edge case.  I'm fine with leaving this as-is for now.  Thanks for helping me understand guys

PR #334, opened 2019-03-27
UIOR-140 location quantities validation error message

AndreiKandratsenka:
## Purpose
 Providing warning messages according to the business expectation
 https://issues.folio.org/browse/UIOR-140

## Approach
 Changed values for warning keys

## Pre-Merge Checklist


translations/ui-orders/en.json
```diff
@@ -402,9 +402,9 @@
   "linesLimit.reasonText": "This would exceed the maximum number of purchase order lines permitted by system settings.",
   "location.button.addLocation": "Add Location",
   "location.nameCode": "Name (Code)",
-  "location.quantityElectronic.exceeds": "Quantity Electronic exceeds Cost Details quantity",
+  "location.quantityElectronic.not.match": "PO Line electronic quantity and Locations electronic quantity do not match",
   "location.quantityElectronic": "Quantity Electronic",
-  "location.quantityPhysical.exceeds": "Quantity Physical exceeds Cost Details quantity",
+  "location.quantityPhysical.not.match": "PO Line physical quantity and Locations physical quantity do not match",
```
aliaksei-chumakou: due to established pattern it should be `location.quantityPhysical.notMatch` IMHO

src/components/Utils/formFieldValidators.js
```diff
@@ -0,0 +1,26 @@
...
+  .map(item => item(...args))
+  .find(item => item);
+
+const reduceLocations = (locations, propName) => {
+  const reducer = (accumulator, d) => accumulator + (d[propName] || 0);
+
+  return locations.reduce(reducer, 0);
+};
+
+export const positiveNumbers = value => (value && value < 0
```
aliaksei-chumakou: do you really need `(` here?

AndreiKandratsenka: yep, EsLint asks about it

translations/ui-orders/en.json
AndreiKandratsenka: fixed

src/components/Utils/formFieldValidators.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import { get } from 'lodash';
+
+export const combineValidators = (validators) => (...args) => validators
```
aliaksei-chumakou: Are you sure we need this? redux-form supports already multiple validation functions, https://redux-form.com/7.4.2/examples/fieldlevelvalidation/

src/components/Utils/formFieldValidators.js
aliaksei-chumakou: `positiveNumbers`, but you're filtering only negative numbers. It doesn't mean that number is positive, it might be empty string, null, etc. I'm not sure that we need to break validation function on several pieces, maybe it makes sense to extend `locationQuantityMatchToDetails` and reduce number of messages, since they still fit

src/components/Utils/formFieldValidators.js
```diff
@@ -0,0 +1,26 @@
...
+  .map(item => item(...args))
+  .find(item => item);
+
+const reduceLocations = (locations, propName) => {
```
NikitaSedyx: name doesn't say a lot, what result of the reduce? it can be anything

NikitaSedyx: I see you just moved code, but please rename

src/components/Utils/formFieldValidators.js
AndreiKandratsenka: ok, but I've thought about combining validators in one place and reuse in the future, if we add check on negative values in `locationQuantityMatchToDetails` it would be validator for singular place, but I need to get used to the project and understand common approach to development on it

src/components/Utils/formFieldValidators.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import { get } from 'lodash';
+
+export const combineValidators = (validators) => (...args) => validators
+  .map(item => item(...args))
+  .find(item => item);
```
NikitaSedyx: ah?

AndreiKandratsenka: removed based on @aliaksei-chumakou comment

src/components/Utils/formFieldValidators.js
aliaksei-chumakou: basically, it's not a common approach, it's handling this code - name doesn't correspond the action.

AndreiKandratsenka: Does you comment mean that I should upgrade the method for checking positive number with additional functionality to check also empty string, null, etc with renaming it< or you would prefer left the code as it exist and simply rename the translation constants values?

aliaksei-chumakou: Read my first comment as two parts: firstly, the name of function is slightly wrong. 
secondly, I suggested to not implement separate validators and mix some specific conditions for location quantities in one validator. 
I think the first one is pretty good candidate to fix, if you agree with me. The second point is up to you

src/components/Utils/formFieldValidators.js
aliaksei-chumakou: @NikitaSedyx  criticize - offer )

NikitaSedyx: oximiron?)

I assume it's `getAllLocationsCount`?

src/components/Utils/formFieldValidators.js
AndreiKandratsenka: Thanks, good point to revise redux-form. Fixed

src/components/Utils/formFieldValidators.js
AndreiKandratsenka: I've decided to leave separate validators and place all of them in one file

aliaksei-chumakou: I don't see separate validators, it's still one validator per field. Also I don't think that current solution satisfies answer from Dennis, that you've asked - please check his comment https://issues.folio.org/browse/UIOR-140?focusedCommentId=47554&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-47554.
So as a solution _now_ it might be a good idea to create a separate validator for required condition (actually we have something already in components/Utils/Validate.js, but w/o translation), and re-think your validators to use them as chain in field's validation (`validate={[v1,v2]}`}

AndreiKandratsenka: The discussion is becoming too long and I'd like to suggest move it in face to face conversation. It might be better to close the PR and open the new one with changing warning messages and additional validation negative values for the field and after accepting common solution on validators return again in this place?

AndreiKandratsenka:
redundant logic

PR #337, opened 2019-03-27
UIOR-188 Filtering by status doesn't work for order results

NikitaSedyx:
## Purpose

Fix filtering by status on orders page
https://issues.folio.org/browse/UIOR-188

## Approach

Used correct property and values

## Pre-Merge Checklist


src/components/Utils/FilterConfig.js
```diff
@@ -7,10 +13,10 @@ const Filters = () => {
...
     },
     {
-      label: 'Approval Status',
-      name: 'approvalStatus',
-      cql: 'approvalStatus',
-      values: ['Approved', 'Pending', 'On hold', 'Not approved'],
+      label: <FormattedMessage id="ui-orders.order.workflow_status" />,
+      name: 'workflowStatus',
+      cql: 'workflowStatus',
+      values: values(WORKFLOW_STATUS),
```
aliaksei-chumakou: `Object.values(WORKFLOW_STATUS)` ? and no import of 3rd party, what do you think?

NikitaSedyx: do we have polyfill for  `Object.values`? as it doesn't support all browsers

lodash is already included into the project, why not to use it

aliaksei-chumakou: according to the https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values it's not supported only in IE, (OK in edge), I don't think we support IE actually, but for sure it's safer to use lodash's thing

NikitaSedyx: wow, safari is green, as I remember it didn't previously

PR #343, opened 2019-03-28
UIOR-140-physical quantities validation error message

AndreiKandratsenka:
## Purpose
provide expected warning messages
https://issues.folio.org/browse/UIOR-140

## Approach
changed warning messages values

## Pre-Merge Checklist


src/components/POLine/Location/LocationForm.js
```diff
@@ -23,13 +23,21 @@ const reduceLocations = (locations, propName) => {
   return locations.reduce(reducer, 0);
 };
 
+const validateShouldBePositive = (value) => {
```
aliaksei-chumakou: you're still have misleading name of function (`positive` in it). `0` is neither negative or positive. So maybe it's better to name it as `validateNotNegative`, what do you think?

AndreiKandratsenka: fixed

src/components/POLine/Location/LocationForm.js
```diff
@@ -23,13 +23,21 @@ const reduceLocations = (locations, propName) => {
   return locations.reduce(reducer, 0);
 };
 
+const validateNotNegative = (value) => {
+  const parsedValue = parseInt(value, 10);
+
+  return Number.isNaN(parsedValue) || parsedValue < 0
```
aliaksei-chumakou: I'm not sure that we should forbid such falsy values as `''`, `undefined` or `null`. There are no clue if the field is required or not.

src/components/POLine/Location/LocationForm.js
```diff
@@ -23,13 +23,21 @@ const reduceLocations = (locations, propName) => {
...
+};
+
 const validateQuantityPhysical = (value, { cost, locations = [] }) => {
   const allLocationsQuantity = reduceLocations(locations, 'quantityPhysical');
   const overallLineQuantity = get(cost, 'quantityPhysical', 0);
 
   return allLocationsQuantity <= overallLineQuantity
     ? undefined
-    : <FormattedMessage id="ui-orders.location.quantityPhysical.exceeds" />;
+    : <FormattedMessage id="ui-orders.location.quantityPhysical.notMatch" />;
```
NikitaSedyx: @aliaksei-chumakou somewhere mentioned that it's better to add new translation than edit existing one, as I got it correctly

AndreiKandratsenka: yep, but I've realized that thing like don't change only value for translation, for preventing skipping the new value with old key

src/components/POLine/Location/LocationForm.js
AndreiKandratsenka: but if we expect some number more than -1, it means for me we should prevent setting undefined, null and empty string isn't it?

src/components/POLine/POLineForm.js
```diff
@@ -72,7 +72,7 @@ class POLineForm extends Component {
         [ACCORDION_ID.itemDetails]: false,
         [ACCORDION_ID.physical]: false,
         [ACCORDION_ID.fundDistribution]: false,
-        [ACCORDION_ID.location]: false,
+        [ACCORDION_ID.location]: true,
```
aliaksei-chumakou: I don't think this should be the case - always opened accordion with Locations.

src/components/POLine/Location/LocationForm.js
aliaksei-chumakou: No, you're wrong, user might want to clear the input and it will be not valid per your logic

aliaksei-chumakou: just stick to atomic validator you've suggested before - the should wait `Number` type value and check simple conditions, just to make possible to stack them in any order you want. So for `required` - a separate validator, and it's not the case for locations' quantities fields.

AndreiKandratsenka: let's look at the case we want validate for not negative value only, and this case without any additional validation for imputed value like '' and etc. we won't give a warning, because are waiting only Number. What should we do this case? And  related with the current piece of code I've added conditional validation according to the order format.

src/components/POLine/POLineForm.js
AndreiKandratsenka: but locations became required field and we should provide at least a one for setting

src/components/POLine/POLineDetails/FieldOrderFormat.js
```diff
@@ -41,17 +42,25 @@ class FieldOrderFormat extends Component {
     dispatch(change('cost.quantityElectronic', ''));
     dispatch(change('cost.listUnitPriceElectronic', ''));
     dispatch(change('cost.listUnitPrice', ''));
+    dispatch(change('locations', [{}]));
```
aliaksei-chumakou: I'm not sure if it's needed. Previous option I'd like better, it was something like `initialValues.locations = [{}]`.

AndreiKandratsenka: but previous implementation doesn't allow cleaning locations field after changing order format

src/components/POLine/Location/LocationForm.js
```diff
@@ -16,81 +16,98 @@ import {
...
+          </Col>
+          <Col xs={3}>
+            <Field
+              component={TextField}
+              label={<FormattedMessage id="ui-orders.location.quantityElectronic" />}
+              name={`${field}.quantityElectronic`}
+              parse={parseQuantity}
+              type="number"
+              disabled={!isNeedElValidation}
+              validate={isNeedElValidation ? [validateNotNegative, validateQuantityElectronic] : noop}
```
aliaksei-chumakou: what do you think of leaving origin code? it worked as expected and tested - just to check if numbers of quantities matches. It was tested, at least manually

AndreiKandratsenka: related with disabled property we need to prevent enter quantity in field doesn't related to the order format, as you know I've harried to ask Dennis about not negative values for quantity and now have to check it. Any case I need to include this check to code in separate method or I'll need to extends existing methods. I'd appreciate to get you vision on this issue

src/components/POLine/Location/LocationForm.js
```diff
@@ -17,28 +17,36 @@ import {
...
+    : <FormattedMessage id="ui-orders.location.quantityPhysical.notMatch" />;
 };
 
 const validateQuantityElectronic = (value, { cost, locations = [] }) => {
-  const allLocationsQuantity = reduceLocations(locations, 'quantityElectronic');
+  const allLocationsQuantity = getTotalLocationsQuantities(locations, 'quantityElectronic');
   const overallLineQuantity = get(cost, 'quantityElectronic', 0);
 
-  return allLocationsQuantity <= overallLineQuantity
+  if (value < 0) {
```
NikitaSedyx: looks like copy paste

AndreiKandratsenka: exclude into separate method

src/components/POLine/Location/LocationForm.js
AndreiKandratsenka: yep, as we agreed I left original code, but added not negative validation in separate method, as @NikitaSedyx mentioned there was a copy&past

test/bigtest/interactors/layer-po-line.js
```diff
@@ -0,0 +1,9 @@
+import {
+  interactor,
+  text,
+} from '@bigtest/interactor';
+
+export default interactor(class LayerPOLine {
```
aliaksei-chumakou: I don't think we need this, it's the same as LineEditPage

test/bigtest/tests/layer-po-line-test.js
```diff
@@ -0,0 +1,67 @@
+import { beforeEach, describe, it } from '@bigtest/mocha';
```
aliaksei-chumakou: `layer-po-line-test.js` - we don't test components. treat it like e2e tests

AndreiKandratsenka: as I don't understand exactly how to calculate test coverage I decided to create new test file. After your comment I moved my test cases into existing file

test/bigtest/interactors/line-edit-page.js
```diff
@@ -77,6 +77,21 @@ import { ACCORDION_ID } from '../../../src/components/POLine/const';
...
 
+@interactor class QuantityLocationElectronic {
+  static defaultScope = '[name$="].quantityElectronic"]';
+}
+
+@interactor class QuantityLocationPhysical {
+  static defaultScope = '[name$="].quantityPhysical"]';
+}
+
+@interactor class Locations {
```
aliaksei-chumakou: I think `LocationAccordion` is the same thing, no need to implement a dup

test/bigtest/tests/layer-po-line-test.js
aliaksei-chumakou: you don't need to calculate test coverage, it's done on SonarCloud. https://github.com/folio-org/ui-orders/pull/343/checks

PR #353, opened 2019-04-03
UIOR-192-added required attribute for Item Details title field

AndreiKandratsenka:
## Purpose
It needs to mark Item Details Title field as required
https://issues.folio.org/browse/UIOR-192

## Approach
added required attribute and validator for empty value for field

## Pre-Merge Checklist


test/bigtest/tests/line-edit-test.js
```diff
@@ -195,4 +195,29 @@ describe('Line edit test', () => {
       });
     });
   });
+
+  describe('Check existing Item Details Title positive case', () => {
+    const EMPTY_STRING = '';
```
NikitaSedyx: you defined constant for empty string, but below you used just `''` :)

IMO constant is not required here

test/bigtest/tests/line-edit-test.js
```diff
@@ -195,4 +195,29 @@ describe('Line edit test', () => {
...
+      await lineEditPage.itemDetailsAccordion.inputTitle('TEST_VALUE');
+      await lineEditPage.updateLineButton.click();
+    });
+
+    it('Shouldn\'t provide any warning message', () => {
+      expect(lineEditPage.itemDetailsAccordion.errorTitle).to.be.equal(EMPTY_STRING);
+    });
+  });
+
+  describe('Check existing Item Details Title negative case', () => {
```
NikitaSedyx: nit: sounds not good to me `'Check existing Should provide title warning`

test/bigtest/interactors/line-edit-page.js
```diff
@@ -88,6 +90,12 @@ import { ACCORDION_ID } from '../../../src/components/POLine/const';
   funds = collection('[data-test-repeatable-field] [class*=repeatableFieldItem---]');
 }
 
+@interactor class ItemDetailsAccordion {
+  static defaultScope = '#itemDetails';
+  inputTitle = fillable('[name="title"]');
```
NikitaSedyx: `'[name="title"]'` can be a constant

test/bigtest/tests/line-edit-test.js
AndreiKandratsenka: will fix next commit

test/bigtest/interactors/line-edit-page.js
AndreiKandratsenka: Yes, I agree with you, but which place for constant do you see? I see two options inside and outside the interactor and both of them I don't like. Inside interactor we'll need to access the constant like a this. ..., if we move it outside the one it would like we need to do the same for all selectors

test/bigtest/tests/line-edit-test.js
AndreiKandratsenka: I would prefer using constant an in next commit I'll move it on common layer

test/bigtest/interactors/line-edit-page.js
NikitaSedyx: as you said before you like constants, here you used `'[name="title"]'` string twice

I don't see any problems to have it outside, I can't forbid you to update all other places if you want :) but it's not required

test/bigtest/tests/line-edit-test.js
NikitaSedyx: up to you

test/bigtest/tests/line-edit-test.js
```diff
@@ -195,4 +195,29 @@ describe('Line edit test', () => {
...
+
+  describe('Check existing Item Details Title positive case', () => {
+    const EMPTY_STRING = '';
+
+    beforeEach(async function () {
+      await lineEditPage.itemDetailsAccordion.inputTitle('TEST_VALUE');
+      await lineEditPage.updateLineButton.click();
+    });
+
+    it('Shouldn\'t provide any warning message', () => {
```
aliaksei-chumakou: hmm, guys, I was thinking that if we don't use `should` then the proper names could be like `it('doesn't provide any warnings')`

test/bigtest/tests/line-edit-test.js
aliaksei-chumakou: What do you think of changing the name to more appropriate like `EMPTY_ERROR_TITLE` or something?

AndreiKandratsenka: this case we create constant only for using once and it might be better to leave simple ''.

test/bigtest/interactors/line-edit-page.js
AndreiKandratsenka: fixed

test/bigtest/tests/line-edit-test.js
AndreiKandratsenka: fixed

PR #363, opened 2019-04-10
UIOR-197 limited check-in dropdown list item statuses

AndreiKandratsenka:
## Purpose
need to leave only "in Process" status in dropdown
https://issues.folio.org/browse/UIOR-197

## Approach
created new array for available statuses in dropdown

## Pre-Merge Checklist


src/components/CheckIn/ItemsListModal.js
```diff
@@ -67,10 +71,10 @@ const ItemsListModal = ({
       <div className={css.fieldWrapper}>
         <Select
           fullWidth
-          defaultValue={ITEM_STATUS.inProcess}
+          defaultValue={STATUS_IN_PROCESS}
```
aliaksei-chumakou: Actually it looks like we don't need to use `defaultValue`. Status is set on piece here https://github.com/folio-org/ui-orders/blob/master/src/components/CheckIn/CheckInDetails.js#L59

aliaksei-chumakou: Also it will be a great thing if you could leverage an item status select into a separate component, since we have the same field in Receiving

NikitaSedyx: but yesterday we approved change for `defaultValue` to use inProgress

AndreiKandratsenka: Thank you for attentively explanation. fixed.

AndreiKandratsenka: @NikitaSedyx, yes of course, but by logic we set default value for each items in componentDidMount other component and creating default value here does noting

src/components/SelectItemStatus/SelectItemStatus.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Select } from '@folio/stripes-components';
+import { FormattedMessage } from 'react-intl';
+import { ITEM_STATUS } from './const';
+import css from '../CheckIn/ItemsListModal.css';
```
aliaksei-chumakou: I don't think this is the right way to import styles. AFAIK, this is css-modules and css is tightly coupled to Component

AndreiKandratsenka: fixed

src/common/constants/itemStatuses.js
```diff
@@ -0,0 +1,11 @@
...
+export const STATUS_ON_ORDER = 'onOrder';
+export const STATUS_RECEIVED = 'received';
+export const ITEM_STATUS = {
+  [STATUS_IN_PROCESS]: 'In process',
+  [STATUS_ON_ORDER]: 'On order',
+  [STATUS_RECEIVED]: 'Received',
+};
+
+export const CHECK_IN_ITEM_STATUSES = [STATUS_IN_PROCESS];
+export const RECEIVE_ITEM_STATUSES = Object.keys(ITEM_STATUS);
```
aliaksei-chumakou: Status options are the same for receive and check-in, it's the same field, that's why I've asked to leverage it to a separate component

src/components/Receiving/LineDetails.js
```diff
@@ -61,22 +65,11 @@ const LineDetails = ({
...
-            >
-              {(message) => <option value={ITEM_STATUS[key]}>{message}</option>}
-            </FormattedMessage>
-          ))}
-        </Select>
-      </div>
+      <SelectItemStatus
+        statuses={RECEIVE_ITEM_STATUSES}
+        value={ITEM_STATUS[STATUS_IN_PROCESS]}
+        selectClass={css.itemStatusField}
```
aliaksei-chumakou: what is the purpose of passing this class in every place? I think it should be managed inside of component

AndreiKandratsenka: This property is passed only in one place, unfortunately this code was in previous version and I'm not sure to delete it.

src/common/constants/itemStatuses.js
AndreiKandratsenka: but what approach could I use for limiting list for check-in till 1?

src/components/Receiving/LineDetails.js
AndreiKandratsenka: moved the property in common component

src/common/constants/itemStatuses.js
aliaksei-chumakou: I mean `CHECK_IN_ITEM_STATUSES` is the same as `RECEIVE_ITEM_STATUSES` from business perspective - it's the same field. So no need to create several vars

src/components/SelectItemStatus/SelectItemStatus.css
```diff
@@ -0,0 +1,6 @@
+.fieldWrapper {
+  margin-top: 1rem;
+}
+.itemStatusField {
+  width: 90px;
+}
```
aliaksei-chumakou: no line at the end, please check your IDE setup

src/common/constants/itemStatuses.js
AndreiKandratsenka: yes, but for check-in flow we need to provide only one item list...

aliaksei-chumakou: It's not related to Check-in flow. It's related to Item Status field, so it should have the same option list from 1 option

AndreiKandratsenka: I need to ask one more time, does it mean that now we need to provide item status list contains 1 item for both part check-in and receiving?

PR #506, opened 2019-07-04
UIOR-246 automatically select the piece

Feboll:
## Purpose

When adding a barcode to a piece during receiving, automatically select that piece for receipt.
https://issues.folio.org/browse/UIOR-246

## Approach

Select a piece if the barcode has been changed.

## Pre-Merge Checklist

### DIFF
--- a/src/components/Receiving/ItemDetails.js
+++ b/src/components/Receiving/ItemDetails.js
@@ -83,7 +83,10 @@ class ItemDetails extends Component {
       const piece = lineItems[item.poLineId].filter(el => el.id === item.id)[0];
 
       piece.isSelected = !piece.isSelected;
-      allChecked[item.poLineId] = false;
+
+      const isAllChecked = lineItems[item.poLineId].find(line => !line.isSelected || line.isSelected === false);
+
+      allChecked[item.poLineId] = !isAllChecked;
       allChecked.reviewDetails = false;
 
       return {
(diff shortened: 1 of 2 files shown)


src/components/Receiving/ItemDetails.js
```diff
@@ -173,17 +173,24 @@ class ItemDetails extends Component {
...
-    this.setState(({ lineItems }) => {
+    this.setState(({ allChecked, lineItems }) => {
       const updatedLineItems = { ...lineItems };
+      const allCheckedItems = { ...allChecked };
+
       const selectedItem = updatedLineItems[item.poLineId].filter(el => el.id === item.id)[0];
 
       selectedItem[key] = value;
+      selectedItem.isSelected = key === 'barcode' && selectedItem[key].length > 0;
+      const isAllChecked = updatedLineItems[item.poLineId].find(line => !line.isSelected || line.isSelected === false);
```
aliaksei-chumakou: @Feboll Have you checked the option of using `toggleItem(item)`?

Feboll: Yes, I checked. I think this solution is simpler, because in the toggleItem we again find the selected Item and call setState again. I also need to check if I chose this item earlier or not. This is correct for toggleAll.

aliaksei-chumakou: I don't think that we need to duplicate item selecting logic. Could you please refactor this place and leverage select item logic in a function, as well as remove `selectedItem.isSelected = key === 'barcode'` from common `onChangeField` ? it  could be `changeBarcode()` callback, for example.

NikitaSedyx: agreed with @aliaksei-chumakou

PR #508, opened 2019-07-05
UIOR-246 automatically select the piece

Feboll:
## Purpose

When adding a barcode to a piece during receiving, automatically select that piece for receipt.
https://issues.folio.org/browse/UIOR-246

## Approach

Select a piece if the barcode has been changed.

## Pre-Merge Checklist

### DIFF
--- a/src/components/Receiving/ItemDetails.js
+++ b/src/components/Receiving/ItemDetails.js
@@ -182,9 +182,7 @@ class ItemDetails extends Component {
 
       selectedItem[key] = value;
 
-      if (value && key === 'barcode') {
-        if (!item.isSelected && value.length > 0) this.toggleItem(item);
-      }
+      if (value && key === 'barcode' && !item.isSelected) this.toggleItem(item);
 
       return {
         lineItems: updatedLineItems,
--- a/src/components/Receiving/ItemDetails.js
+++ b/src/components/Receiving/ItemDetails.js
@@ -184,7 +184,6 @@ class ItemDetails extends Component {
 
       if (value && key === 'barcode') {
         if (!item.isSelected && value.length > 0) this.toggleItem(item);
-        if (item.isSelected && value.length === 0) this.toggleItem(item);
       }
 
       return {
--- a/src/components/Receiving/ItemDetails.js
+++ b/src/components/Receiving/ItemDetails.js
@@ -182,16 +182,15 @@ class ItemDetails extends Component {
 
       selectedItem[key] = value;
 
+      if (key === 'barcode') {
+        if (!item.isSelected && value.length > 0) this.toggleItem(item);
+        if (item.isSelected && value.length === 0) this.toggleItem(item);
+      }
+
       return {
         lineItems: updatedLineItems,
       };
     });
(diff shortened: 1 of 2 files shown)


src/components/Receiving/ItemDetails.js
```diff
@@ -83,15 +83,18 @@ class ItemDetails extends Component {
       const piece = lineItems[item.poLineId].filter(el => el.id === item.id)[0];
 
       piece.isSelected = !piece.isSelected;
-      allChecked[item.poLineId] = false;
+
+      const isAllChecked = lineItems[item.poLineId].find(line => !line.isSelected || line.isSelected === false);
```
aliaksei-chumakou: it looks like `!line.isSelected` covers the second condition also, does it make sense to include both?

aliaksei-chumakou: Also naming looks weird, you have a var `isAllChecked`, but it will have `false` if all records are selected.

src/components/Receiving/LineDetails.js
```diff
@@ -23,6 +23,7 @@ const LineDetails = ({
   lineItems,
   locationsOptions,
   onChangeField,
+  onChangeBarcode,
```
aliaksei-chumakou: does it make sense to pass that prop? IMHO, it could live here, within the field. It has access to all required props, as far as I see. It's arguable though, maybe additional input from other guys is required , cc @AndreiShumski2 @NikitaSedyx

NikitaSedyx: 100%

src/components/Receiving/ItemDetails.js
```diff
@@ -179,11 +182,16 @@ class ItemDetails extends Component {
 
       selectedItem[key] = value;
 
+      if (key === 'barcode') {
+        if (!item.isSelected && value.length > 0) this.toggleItem(item);
+        if (item.isSelected && value.length === 0) this.toggleItem(item);
```
aliaksei-chumakou: I think there might be situations when value could came as `null` or `undefined`. Don't you think so?

aliaksei-chumakou: also I've checked the ticket and found no requirement to clear selection

Feboll: I don't think so. We don't get these values ​​from elements with the values and if we don't sent manually, but for security, I will add one more condition

src/components/Receiving/ItemDetails.js
```diff
@@ -179,11 +182,15 @@ class ItemDetails extends Component {
 
       selectedItem[key] = value;
 
+      if (value && key === 'barcode') {
+        if (!item.isSelected && value.length > 0) this.toggleItem(item);
```
aliaksei-chumakou: `if` nested into `if`... hmm.... could it be simplified? ... hmmm... who knows...

Feboll: you're right. I updated the condition

NikitaSedyx:
:+1:

PR #548, opened 2019-08-12
UIOR-235 Settings: require approval for orders to be opened

Feboll:
## Purpose

Some institutions require that orders are approved before they can be placed.
https://issues.folio.org/browse/UIOR-235

## Approach

Add new page with approval settings

## Pre-Merge Checklist

### DIFF
--- a/src/common/utils/getOrderApprovalsSetting.js
+++ b/src/common/utils/getOrderApprovalsSetting.js
@@ -2,9 +2,6 @@ import { get } from 'lodash';
 // eslint-disable-next-line import/prefer-default-export
 export const getOrderApprovalsSetting = (configs) => {
   let orderApprovalsSetting = get(configs, [0, 'value'], '{}');
-  const config = {
-    isApprovalRequired: false,
-  };
 
   try {
     orderApprovalsSetting = JSON.parse(orderApprovalsSetting);
@@ -12,7 +9,5 @@ export const getOrderApprovalsSetting = (configs) => {
     orderApprovalsSetting = {};
   }
 
-  Object.assign(config, orderApprovalsSetting);
-
-  return config;
+  return orderApprovalsSetting;
 };
--- a/src/common/utils/getOrderApprovalsSetting.js
+++ b/src/common/utils/getOrderApprovalsSetting.js
@@ -1,6 +1,6 @@
 import { get } from 'lodash';
-
-const getOrderNumberSetting = (configs) => {
+// eslint-disable-next-line import/prefer-default-export
+export const getOrderApprovalsSetting = (configs) => {
   let orderApprovalsSetting = get(configs, [0, 'value'], '{}');
   const config = {
     isApprovalRequired: false,
@@ -16,5 +16,3 @@ const getOrderNumberSetting = (configs) => {
 
   return config;
 };
-
-export default getOrderNumberSetting;
--- a/src/settings/OrderApprovals/OrderApprovals.css
+++ b/src/settings/OrderApprovals/OrderApprovals.css
@@ -1,0 +1,8 @@
+.formWrapper {
+  width: 100%;
+}
+
+.formWrapper > div,
+.formWrapper > div > form {
+  height: 100%;
+}
--- a/src/settings/OrderApprovals/OrderApprovals.css
+++ b/src/settings/OrderApprovals/OrderApprovals.css
@@ -1,8 +1,0 @@
-.formWrapper {
-  width: 100%;
-}
-
-.formWrapper > div,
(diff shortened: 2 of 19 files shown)


src/settings/OrderApprovals/OrderApprovals.css
```diff
@@ -0,0 +1,8 @@
+.formWrapper {
+  width: 100%;
```
aliaksei-chumakou: I don't feel like these are required styles. Could you please elaborate more on the problem you're solving here?

Feboll: This need for stretch the block in width. without this style it looks like this
image

src/settings/OrderApprovals/OrderApprovals.js
```diff
@@ -0,0 +1,51 @@
...
+  render() {
+    const { label } = this.props;
+
+    return (
+      <div
+        data-test-order-settings-order-approvals
+        className={css.formWrapper}
+      >
+        <this.configManager
+          configName="approvals"
```
AndreiShumski2: could be constant

src/common/utils/getOrderApprovalsSetting.js
```diff
@@ -0,0 +1,18 @@
...
+    isApprovalRequired: false,
+  };
+
+  try {
+    orderApprovalsSetting = JSON.parse(orderApprovalsSetting);
+  } catch (e) {
+    orderApprovalsSetting = {};
+  }
+
+  Object.assign(config, orderApprovalsSetting);
```
NikitaSedyx: I think it's redundant, it's enough to handle it in try catch without default values object

src/settings/OrderApprovals/OrderApprovals.css
aliaksei-chumakou: I think you can achieve the same result with layout, without introducing extra specific styles 
https://ux.folio.org/storybook/?selectedKind=LayoutGrid&selectedStory=Basic%20Usage&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel
BTW, please look at another places with settings, how it's handled there

src/settings/OrderApprovals/OrderApprovals.js
```diff
@@ -0,0 +1,51 @@
...
+import { ConfigManager } from '@folio/stripes/smart-components';
+
+import {
+  MODULE_ORDERS,
+} from '../../components/Utils/const';
+import { getOrderApprovalsSetting } from '../../common/utils/getOrderApprovalsSetting';
+import OrderApprovalsForm from './OrderApprovalsForm';
+import css from './OrderApprovals.css';
+
+class OrderApprovals extends Component {
```
NikitaSedyx: seems it should be functional component, but `props.stripes.connect`

Feboll: I think we can't change to stateless component, because then this form does't work.

src/settings/OrderApprovals/OrderApprovals.js
```diff
@@ -1,51 +1,45 @@
...
-        </this.configManager>
-      </div>
+        </div>
+      </this.configManager>
     );
   }
 }
 
+OrderApprovals.propTypes = {
+  label: PropTypes.object.isRequired,
```
aliaksei-chumakou: I don't think `label` is an object. Please find a proper type

Feboll: This is an object, if you trust `typeof`. but if be more precise this is a react element. Do you think I need to change to `PropTypes.element`

aliaksei-chumakou: PropTypes.object gives you nothing in this case, it's like typescripts's `any`.

aliaksei-chumakou: there is a docs regarding proptypes, please check and find the proper one https://reactjs.org/docs/typechecking-with-proptypes.html

Feboll: Ok I change to `PropTypes.element`

aliaksei-chumakou: why do you want to limit this to an element? I've mentioned a `string`, it still don't fit.

NikitaSedyx: why not node?

Feboll: I think node is correct too, `ConfigManager` is waiting for this

PR #608, opened 2019-10-16
UIOR-405: Fix bug with reassign template when editing new PO

Vladyslav-Velytskyi:


src/components/PurchaseOrder/POForm.js
```diff
@@ -193,6 +193,7 @@ class POForm extends Component {
     const vendors = get(parentResources, 'vendors.records', [])
       .filter(vendor => vendor.isVendor && vendor.status === ORGANIZATION_STATUS_ACTIVE);
     const orderTemplates = getOrderTemplatesForSelect(parentResources);
+    const poLinesLength = get(initialValues, 'compositePoLines', '').length;
```
aliaksei-chumakou: but it's an array, why do you use empty string as a default value?

Vladyslav-Velytskyi: Thanks, it was misprint, i'll fix it

aliaksei-chumakou:
@Vladyslav-Velytskyi  you can run tests locally by `yarn test` in `ui-orders`

PR #741, opened 2020-01-22
UIOR-492: Interface migration to new versions of circulation and inventory

Baroquem:
https://issues.folio.org/browse/UIOR-492

Note that I'm not 100% confident that I've caught all the places where the new rules about item.status need to be updated; this is my first time working on the ui-orders code. As far as I can tell, it seems to be working.


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change history for ui-orders
 
+## (IN PROGRESS)
```
aliaksei-chumakou: too much _(in progress)_ :smiley_cat: 1.9.0 is already in progress below)

PR #839, opened 2020-04-22
FundFilterContainer

AndreiKononov:
## Pre-Merge Checklist
- [ ] I've added appropriate record to the CHANGELOG.md
  - [x] There are no breaking changes in this PR.


src/common/FundFilter/FundFilterContainer.js
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <FundFilter
+      {...rest}
+      funds={options}
+    />
+  );
+};
+
+FundFilterContainer.propTypes = {
+  funds: PropTypes.arrayOf(PropTypes.object),
```
aliaksei-chumakou, line 20: actually. that was the point - to not use passing prop and load it in the container here. Please look at the `resources` prop to pass funds from back-end to the presentation component

src/common/FundFilter/FundFilter.js
```diff
@@ -1,25 +1,20 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
 import { SelectionFilter } from '@folio/stripes-acq-components';
-
-import {
-  getFundOptions,
-} from '../utils';
-import { fundsShape } from '../shapes';
+import { fundShape } from '../shapes';
```
aliaksei-chumakou, line 5: I don't get this change, could you please provide a reason?

AndreiKononov: It happened due to Eslint asked to move line `import PropTypes from 'prop-types';` above than other imports.

AndreiKononov: Not actual after the last commit.

src/common/FundFilter/FundFilterContainer.js
AndreiKononov: Done.
- added `manifest`

- used `resources` prop to pass funds from back-end

- updated `getFundOptions` method regarding components updates

src/common/FundFilter/FundFilter.js
```diff
@@ -1,25 +1,19 @@
...
 
 import { SelectionFilter } from '@folio/stripes-acq-components';
 
-import {
-  getFundOptions,
-} from '../utils';
-import { fundsShape } from '../shapes';
-
 const FundFilter = ({ funds, ...rest }) => {
-  const options = getFundOptions(funds);
```
aliaksei-chumakou, line 11: I actually don't think you need to touch this file, or maybe you have a reason for that?

src/common/utils/getFundOptions.js
```diff
@@ -1,5 +1,8 @@
+import { get } from 'lodash';
+import { DICT_FUNDS } from '@folio/stripes-acq-components';
+
 // eslint-disable-next-line import/prefer-default-export
-export const getFundOptions = (funds = []) => funds.map(fund => ({
+export const getFundOptions = (resources) => get(resources, [DICT_FUNDS, 'records'], []).map(fund => ({
```
aliaksei-chumakou, line 5: what is the idea behind this change?

src/common/FundFilter/FundFilter.js
AndreiKononov: Sorry, but I'd like to clarify. Which file do you mean? `getFundOptions.js` or `FundFilter.js`?

src/common/utils/getFundOptions.js
AndreiKononov: Firstly, to avoid an error:
image;
Secondly, a similar approach was used for the container component which was provided as an example.
image

aliaksei-chumakou: first of all, you shouldn't copy-paste just to make something similar - you should realize what you're doing. so the second point is not valid, please provide why you want to make this here.
regarding error - IMO, when you come to the code base you're not familiar with the best way to not break things is to make as less changes as possible. That's why I don't see any point to change the View component, it was working before. What you can do - is to avoid such error in Container. Obviously it's occurred when you're trying to render presentation component while list of options still loading from the back-end. Maybe it makes sense to not render presentation component while it's happening.

src/common/FundFilter/FundFilter.js
aliaksei-chumakou: both, please look at my another comment for clarification

AndreiKononov: both files were rolled to initial state back

src/common/utils/getFundOptions.js
AndreiKononov: Thanks for the advice, it helped me to resolve the issue of data loading into Container.
Regarding the question, I made it there because I was trying to fix the error and found out this approach in the example. As a result, it helped me to avoid this error.

src/common/FundFilter/FundFilterContainer.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { stripesConnect } from '@folio/stripes/core';
+import { DICT_FUNDS, fundsManifest } from '@folio/stripes-acq-components';
+
+import FundFilter from './FundFilter';
+
+const FundFilterContainer = ({ resources, ...rest }) => {
+  const funds = resources[DICT_FUNDS].records;
```
aliaksei-chumakou, line 10: `resources[DICT_FUNDS]` might be null

aliaksei-chumakou:
mostly looks good now, please sign CLAassistant  by link in comments above. Also it would be great if you can remove loading and passing funds from components that uses FundFilter.

src/common/FundFilter/FundFilterContainer.js
AndreiKononov: Added the optional chaining operator.

AndreiKononov:
> remove loading and passing funds from components that uses FundFilte

AndreiKononov:
> mostly looks good now, please sign CLAassistant by link in comments above. Also it would be great if you can remove loading and passing funds from components that uses FundFilter.

I've already signed to CLAassistant. 
I found out the only component that uses FundFilter component. It's OrderLinesFilters. Will let you know when I will finish with them.

aliaksei-chumakou:
@AndreiKononov Looks good so far, however, could you please add unit test for presentation component? 
for reference: 
https://github.com/folio-org/stripes-acq-components/blob/master/lib/AcqCheckboxFilter/AcqCheckboxFilter.test.js
or any other jest unit tests (.test.js)

src/common/FundFilter/FundFilter.test.js
```diff
@@ -0,0 +1,33 @@
...
+  </IntlProvider>,
+));
+
+describe('FundFilter component', () => {
+  afterEach(cleanup);
+
+  it('should display passed title', () => {
+    const { getByText } = renderFundFilter();
+
+    expect(getByText(filterAccordionTitle)).toBeDefined();
```
aliaksei-chumakou, line 31: IMHO, more tests could be added, for selection is called, etc.

AndreiKononov: Added additional tests

src/common/FundFilter/FundFilterContainer.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { stripesConnect } from '@folio/stripes/core';
+import { DICT_FUNDS, fundsManifest } from '@folio/stripes-acq-components';
```
AndreiShumski2, line 5:
```suggestion
import {
  DICT_FUNDS,
  fundsManifest,
 } from '@folio/stripes-acq-components';
```

PR #840, opened 2020-04-22
UIOR-495 Populate poNumberPrefix & poNumberSuffix fields on order creation

AndreiShumski2:
## Purpose
https://issues.folio.org/browse/UIOR-495

## Pre-Merge Checklist

### DIFF
--- a/src/components/Utils/orderResource.js
+++ b/src/components/Utils/orderResource.js
@@ -15,7 +15,7 @@ const saveOrder = (order, mutator) => {
   return method(order);
 };
 
-export const getFullOrderNumber = ({ poNumberPrefix = '', poNumberSuffix = '', poNumber = '' }) => (
+export const getFullOrderNumber = ({ poNumberPrefix = '', poNumberSuffix = '', poNumber = '' } = {}) => (
   `${poNumberPrefix}${poNumber}${poNumberSuffix}`.trim()
 );
 
--- a/src/components/Utils/orderResource.js
+++ b/src/components/Utils/orderResource.js
@@ -6,8 +6,6 @@ const saveOrder = (order, mutator) => {
   delete order.createdByName;
   delete order.assignedToUser;
   delete order.vendorName;
-  delete order.numberPrefix;
-  delete order.numberSuffix;
 
   if (order.id) {
     method = mutator.PUT;
@@ -17,8 +15,14 @@ const saveOrder = (order, mutator) => {
   return method(order);
 };
 
+export const getFullOrderNumber = ({ poNumberPrefix = '', poNumberSuffix = '', poNumber = '' }) => (
+  `${poNumberPrefix}${poNumber}${poNumberSuffix}`.trim()
+);
+
 export const updateOrderResource = (order, mutator, changedProps) => {
   const clonedOrder = cloneDeep(order);
+
+  clonedOrder.poNumber = getFullOrderNumber(clonedOrder);
 
   Object.assign(clonedOrder, changedProps);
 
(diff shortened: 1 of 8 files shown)


src/components/Utils/orderResource.js
```diff
@@ -17,26 +15,33 @@ const saveOrder = (order, mutator) => {
...
   return saveOrder(clonedOrder, mutator);
 };
 
 export const createOrderResource = (order, mutator) => {
   const clonedOrder = cloneDeep(order);
-  const { numberPrefix = '', numberSuffix = '', poNumber = '' } = clonedOrder;
-  const fullOrderNumber = `${numberPrefix}${poNumber}${numberSuffix}`.trim();
 
-  clonedOrder.poNumber = fullOrderNumber || undefined;
+  clonedOrder.poNumber = getFullOrderNumber(clonedOrder) || undefined;
```
aliaksei-chumakou, line 35: have you checked with the back-end if we still need this?

src/components/Utils/orderResource.js
```diff
@@ -17,26 +15,33 @@ const saveOrder = (order, mutator) => {
   return method(order);
 };
 
+export const getFullOrderNumber = ({ poNumberPrefix = '', poNumberSuffix = '', poNumber = '' }) => (
```
aliaksei-chumakou, line 18: it will throw an exception if null or undefined is passed. Maybe it makes sense to add a default value
```suggestion
export const getFullOrderNumber = ({ poNumberPrefix = '', poNumberSuffix = '', poNumber = '' } = {}) => (
```

src/components/Utils/orderResource.js
AndreiShumski2: The story description provides it: 
image

src/components/Utils/orderResource.js
AndreiShumski2: correct, missed it

src/components/Utils/orderResource.js
```diff
@@ -17,26 +15,33 @@ const saveOrder = (order, mutator) => {
   return method(order);
 };
 
+export const getFullOrderNumber = ({ poNumberPrefix = '', poNumberSuffix = '', poNumber = '' } = {}) => (
```
aliaksei-chumakou, line 18: **FYI, not a required change**
usually brackets are needed if you return a react element. In this very case it should be OK without them.

src/components/Utils/orderResource.js
```diff
@@ -17,26 +15,33 @@ const saveOrder = (order, mutator) => {
...
+  clonedOrder.poNumber = getFullOrderNumber(clonedOrder) || undefined;
 
   return saveOrder(clonedOrder, mutator);
 };
 
 export const cloneOrder = (order, mutator, lines) => {
-  const clonedOrder = omit(order, ['id', 'adjustment', 'metadata', 'poNumber', 'workflowStatus', 'compositePoLines']);
+  const clonedOrder = omit(
+    order,
+    ['id', 'adjustment', 'metadata', 'poNumber', 'poNumberPrefix', 'poNumberSuffix', 'workflowStatus', 'compositePoLines'],
```
aliaksei-chumakou, line 43: What is the reason to skip here suffix and prefix?? To me it looks like a bug

AndreiShumski2: When user clones order PO number generates on BE side (only numbers). So no omitting pref/suf will brake PO number creation. I mean we'll get
`{
    poNumberPrefix: 'FOO',
    poNumber: '1001',
    poNumberSuffix: 'BAR'
}
`
instead of `{
    poNumberPrefix: 'FOO',
    poNumber: 'FOO1001BAR',
    poNumberSuffix: 'BAR'
}
`

aliaksei-chumakou: what makes you think like that? right now suffix and prefix are omitting and if you clone the order `{ poNumberPrefix: 'FOO', poNumber: '1001', poNumberSuffix: 'BAR' }` you'll get `{ poNumber: '1002' }`. So my question was why not to leave prefix and suffix, but now I see that PO will still be generated as `1002` and ui will be broken since it will try to remove suffix and prefix from that generated number

PR #1028, opened 2021-01-14
Create buildNPM.yml

ankitasen-ubmainz:
FOLIO-2913 SPIKE: Evaluate GitHub Actions


funkymalc (approved):
- Can the value of '-Dsonar.projectName' be a var  (similar to -Dsonar.projectKey)? 
- '-Dsonar.sources=.' should be './src'.  Note that this is a configurable option in Jenkins now via the Jenkinsfile with the default defined as './src'.  

Also see comments in https://github.com/folio-org/mod-orders/pull/407

PR #1213, opened 2021-10-19
import babel options via @folio/stripes-cli

zburke:
## Purpose

Import babel-options via `@folio/stripes-cli` in order to retrieve a
correctly-configured options object (suppressing a LOT of error
messaging during testing) without maintaining a local copy.

Without this aligning the options, test output is overrun with
> Though the `"loose"` option was set to "false" in your @babel/preset-env config, it will not be used for `@babel/plugin-proposal-private-property-in-object` since the `"loose"` mode option was set to "true" for `@babel/plugin-proposal-private-methods.`
> The `"loose"` option must be the same for `@babel/plugin-proposal-class-properties`, `@babel/plugin-proposal-private-methods` and `@babel/plugin-proposal-private-property-in-object` (when they are enabled): you can silence this warning by explicitly adding
>```
>    ["@babel/plugin-proposal-private-property-in-object", { "loose": true }]`
>```
> to the `"plugins"` section of your Babel config.

## Approach

Import the options from `@folio/stripes-cli`, which are re-exported directly from `@folio/stripes-webpack`, and therefore are guaranteed to be current. 

#### TODOS and Open Questions

Cleaning up the babel noise revealed quite a lot of noise about prop-types violations, invalid props (probably props objects that are spread and then trickle down to the HTML components), and async react state updates ("Can't perform a React state update on an unmounted component..."), as well as several from RTL about test configuration. Some of that is junk that likely only shows up during testing, but much of it will seriously pollute the console in a dev-env and should be cleaned up if possible. 

And there's one test continues to fail for me, though it appears to have run cleanly in Jenkins 😕 : 
```
 FAIL  src/common/hooks/useHandleOrderUpdateError/useHandleOrderUpdateError.test.js
  ● Test suite failed to run
    Call retries were exceeded
      at ChildProcessWorker.initialize (node_modules/jest-worker/build/workers/ChildProcessWorker.js:193:21)
```


zburke:
Ping: @NikitaSedyx, @AndreiShumski2, @usavkov-epam. If there are no objections, I will go ahead and merge this in a day or two.

zburke:
This PR is likely outdated and this isn't my repository to begin with.

PR #1337, opened 2022-03-25
Github Actions Implementation:Do not merge

ankitasen-ubmainz:


ankitasen-ubmainz:
@NikitaSedyx do you have any updates about this? The PR @zburke merged resolved the issue which was mentioned in [UIOR-997](https://issues.folio.org/browse/UIOR-997), but it gave rise to another one now and the issue is again put in open status. This is the only ui repository remaining which is still not migrated to GA so it is important we solve this.

AndreiShumski2:
@ankitasen-ubmainz, now all unit tests passed successfully

PR #1393, opened 2022-08-01
UIOR-997 do not iterate if locations is null

zburke:
Do not call `list.forEach` if `list` is null. [Line 104 sets `locations`
to a possibly null value](https://github.com/folio-org/ui-orders/blame/8cd095498aca31589d147a32e888073f0d4b72e7/src/components/POLine/Item/ItemForm.js#L101-L116):
```
    const locations = formValues?.locations;
```
and therefore we must validate it before calling methods on it.

This should help unblock #1337.

Refs [UIOR-997](https://issues.folio.org/browse/UIOR-997)

### DIFF
--- a/src/components/POLine/Item/ItemForm.js
+++ b/src/components/POLine/Item/ItemForm.js
@@ -110,7 +110,9 @@ class ItemForm extends Component {
         change('instanceId', inventoryData.instanceId);
       } else {
         change('instanceId', null);
-        locations.forEach((_, i) => change(`locations[${i}].holdingId`, null));
+        if (Array.isArray(locations)) {
+          locations.forEach((_, i) => change(`locations[${i}].holdingId`, null));
+        }
       }
     });
   };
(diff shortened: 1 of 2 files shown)


src/components/POLine/Item/ItemForm.js
```diff
@@ -110,7 +110,9 @@ class ItemForm extends Component {
         change('instanceId', inventoryData.instanceId);
       } else {
         change('instanceId', null);
-        locations.forEach((_, i) => change(`locations[${i}].holdingId`, null));
```
AndreiShumski2, line 113: `locations?.forEach...` should work here as well.
Thanks @zburke for getting into it.

ankitasen-ubmainz (approved):

src/components/POLine/Item/ItemForm.js
zburke: I see that's how it's implemented a few lines later in `setIsPackage` so I will change this line to be consistent with that one. Good thinking; thanks for mentioning it.

usavkov-epam (approved):

NikitaSedyx (approved):

PR #1672, opened 2024-11-12
UIOR-1339 Fix print order action

alb3rtino:
https://folio-org.atlassian.net/browse/UIOR-1339

## Purpose
* Fix broken 'Print order' action

## Approach
* Add missing argument to `getExportData` method

### DIFF
--- a/src/PrintOrder/PrintOrder.js
+++ b/src/PrintOrder/PrintOrder.js
@@ -37,7 +37,7 @@ export const PrintOrderComponent = ({ mutator, order, orderLine, onCancel }) => 
       const { compositePoLines } = order;
       const linesToPrint = orderLine ? [orderLine] : compositePoLines;
       const printData = compositePoLines?.length
-        ? { lines: await getExportData(mutator, linesToPrint, [order], intl) }
+        ? { lines: await getExportData(mutator, linesToPrint, [order], [], intl) }
         : await getOrderPrintData(ky, order);
 
       let poLineTotals = {};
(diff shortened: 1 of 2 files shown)


src/PrintOrder/PrintOrder.js
```diff
@@ -37,7 +37,7 @@ export const PrintOrderComponent = ({ mutator, order, orderLine, onCancel }) =>
       const { compositePoLines } = order;
       const linesToPrint = orderLine ? [orderLine] : compositePoLines;
       const printData = compositePoLines?.length
-        ? { lines: await getExportData(mutator, linesToPrint, [order], intl) }
+        ? { lines: await getExportData(mutator, linesToPrint, [order], [], intl) }
```
NikitaSedyx, line 40: so we don't support custom fields in this print template? (I don't know requirements for custom fields that's why I ask)

alb3rtino: Correct. A requirement was only defined for csv export ([UIOR-1231](https://folio-org.atlassian.net/browse/UIOR-1231)).

NikitaSedyx (approved):

src/PrintOrder/PrintOrder.js
NikitaSedyx: thanks, I wanted to put empty array as well, but wasn't sure about requirements so redirected this ticket to your team :)

SerhiiNosko (approved):

usavkov-epam (approved):

PR #1740, opened 2025-05-27
UIOR-1435: Add `PO_LINE_CONFIG_NAME_PREFIX` constant and use it in the custom field components.

Dmytro-Melnyshyn:
## Description
When multiple instances of the `EditCustomFieldsSettings`/ `EditCustomFieldsRecord`/ `ViewCustomFieldsRecord`/ `ViewCustomFieldsSettings` UI component, each associated with a distinct `entityType`, are used within a single application, then modifying the accordion label in one component results in an unintended change to the label in all other components. This behavior contradicts the expected functionality, causing inconsistency in the labeling of accordion sections based on entityTypes.

The issue arises due to the request sent to the configurations API to store the accordion label, which utilizes a single `configName` for all entityTypes.

Using a single `configName` for multiple entityTypes, leads to a universal change in the accordion label across all instances of the `EditCustomFieldsSettings`/ `EditCustomFieldsRecord`/ `ViewCustomFieldsRecord`/ `ViewCustomFieldsSettings` components, regardless of their associated entityTypes.

## Related PRs
https://github.com/folio-org/stripes-smart-components/pull/1584

## Approach
Use `configNamePrefix` property to extend `configName` for the PO line and leave the PO as is.

## Issues
[UIOR-1435](https://folio-org.atlassian.net/browse/UIOR-1435)

## Screencasts

https://github.com/user-attachments/assets/fa6adc9f-cb8d-4b70-83b1-ac87f53aee48

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Add "Order template categories" field to order template. Refs UIOR-1401.
 * Add Claiming active, Claiming interval fields, Donor information accordion to Template. Refs UIOR-1393.
 * Add ability to hide optional fields in the order template form. Refs UIOR-1394.
+* Add PO_LINE_CONFIG_NAME_PREFIX constant and use it in the custom field components. Fixes UIOR-1435.
 
 ## [8.0.3](https://github.com/folio-org/ui-orders/tree/v8.0.2) (2025-04-10)
 [Full Changelog](https://github.com/folio-org/ui-orders/compare/v8.0.2...v8.0.3)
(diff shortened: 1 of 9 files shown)


OleksandrHladchenko1 (approved):

zburke (approved):
If we leverage `configNamePrefix` in `CustomFieldsSettings` on line 66 and 77 when handling PO Lines, shoudn't we do the same on lines 43 and 53 when handling Orders? Technically, I guess we just need to make sure those values are distinct, and now they are, but changing only one feels wrong somehow.

PR #1866, opened 2026-06-23
UIOR-1519 *BREAKING* Update CQL queries to use the new indices

usavkov-epam:
## Purpose

https://folio-org.atlassian.net/browse/UIOR-1519

Requires https://github.com/folio-org/stripes-acq-components/pull/951

## Approach

- Use `buildMultiOptionCqlQuery` to update CQL with relation modifiers;
- Remove wrapping the filter value in wildcards.
- Update tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 # Change history for ui-orders
 
-## 9.1.0 (IN PROGRESS)
+## 10.0.0 (IN PROGRESS)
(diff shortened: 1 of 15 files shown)


usavkov-epam:
Tests will pass after https://github.com/folio-org/stripes-acq-components/pull/951 is merged

SerhiiNosko (approved):

PR #1870, opened 2026-07-16
UIOR-1157 Deprecate acquisition methods in settings, PO line and templates

slaemmer:
[UIOR-1157](https://folio-org.atlassian.net/browse/UIOR-1157)

## Purpose
Let an Acquisitions admin toggle `deprecated` on an acquisition method in Settings, hide deprecated methods from the PO-line and Order-template dropdowns, keep them visible/suffixed where they already appear, and warn when opening an order whose lines use a deprecated method. Depends on stripes-acq-components (UISACQCOMP-301).

## Approach
Soft enforcement, same pattern as prefix/suffix deprecation: existing records keep working, deprecated methods just stop being offered.

- Settings gets a `Deprecated` checkbox on every method, System ones included. Those are the ones libraries actually want to retire. System names stay read-only (`getReadOnlyFieldsForItem`) since GOBI and data-import match by name; delete stays blocked.
- The PO-line dropdown (`FieldAcquisitionMethod`) hides deprecated methods but keeps the line's saved one, suffixed `(deprecated)`, hiding it would break the record for whoever edits the line next. The template editor shares the field, so it behaves the same. The view just shows the suffix.
- New order from a template whose method got deprecated: the field comes up empty, and since it's required the user has to pick a current one.
- Opening such an order still works. `OpenOrderConfirmationModal` looks up the deprecated methods via the new `useDeprecatedAcqMethods` hook (fetches only while the dialog is mounted) and shows a banner listing them with their PO line numbers; Submit is disabled until the check lands, so it can't be clicked past.
- `acquisition-methods` interface floor raised to `1.1`.
- Label and filter logic lives in stripes-acq-components (`useAcqMethodsOptions`, `getAcqMethodLabel`) so ui-orders and find-po-line render deprecation identically; the local label helper is deleted.

## Screenshots
Deprecation checkboxes exist, can be checked, name of system acquisition methods is immutable
Screenshot from 2026-07-16 10-06-23
Screenshot from 2026-07-16 10-06-51
Librarian is warned when trying to open a order with deprecated acquisition method(s) in po line(s)
Screenshot from 2026-07-16 10-08-02
Deprecated suffix is displayed
Screenshot from 2026-07-16 10-08-24
Previously selected deprecated method can be selected when editing an existing po line
Screenshot from 2026-07-16 10-09-03
Deprecated methods cant be selected when creating a new po line
Screenshot from 2026-07-16 10-10-13

## Pre-Merge Checklist
  - [x] If any API-related changes - okapi interfaces and permissions are reviewed/changed correspondingly
    - `acquisition-methods` required interface floor raised `1.0 to 1.1`
    - The breaking part is the raised interfaces/deps: the `acquisition-methods 1.1` interface requirement and the `stripes-acq-components ~7.2.0` dependency (see the deployment note below for ordering).

The required acq-models, mod-orders, mod-orders-storage, and stripes-acq-components PRs were reviewed & merged.


src/common/hooks/useDeprecatedAcqMethods/useDeprecatedAcqMethods.js
usavkov-epam, line 1: In `useDeprecatedAcqMethods`, instead of calling `useAcqMethods()` which fetches all acquisition methods with `cql.allRecords=1` and then filtering them in memory, consider optimizing this to use CQL-based server-side filtering on the `deprecated` field. 

Currently, the hook:
1. Fetches all methods from the API
2. Filters by `deprecated && linesByMethodId[id]` in `useMemo`

A more efficient approach would query only the deprecated methods needed for the warning banner. Since `acquisition-methods` interface was bumped to v1.1, this should support server-side CQL filtering. This would reduce bandwidth, especially for deployments with many acquisition methods, and align with the pattern of pushing filtering logic to the API layer where possible.

src/common/POLFields/DetailsFields/FieldAcquisitionMethod.js
```diff
@@ -15,7 +16,16 @@ const FieldAcquisitionMethod = ({
   ...props
 }) => {
   const { acqMethods } = useAcqMethods();
-  const acquisitionMethods = useAcqMethodsOptions(acqMethods);
+  const { meta: { initial: initialAcqMethod } } = useField(POL_FORM_FIELDS.acquisitionMethod, {
+    subscription: { initial: true },
+  });
```
usavkov-epam: Looking at the use of `useField` here to extract the initial value — it works, but it registers the field with the form a second time (the first being via `FieldSelectionFinal`). 

Since you only need the initial value for the options logic, consider using `useForm()` instead:

```javascript
const { getFieldState } = useForm();
const initialAcqMethod= getFieldState(POL_FORM_FIELDS.acquisitionMethod)?.initial;
```

This avoids the double registration entirely. getFieldState() is a read-only query into the form's field registry — it doesn't create a subscription or cause re-renders. It's more semantically clear that you're just reading state, not observing field changes, and it's the cleaner pattern when you need a one-time read of a field's metadata.

src/common/hooks/useDeprecatedAcqMethods/useDeprecatedAcqMethods.js
slaemmer: I see the point of pushing filtering logic to the api layer when possible and yes, generally this would be more efficient. 

I decided against it though as `useAcqMethods((` is called throughout the codebase. By the time someone opened an order it would most likely only be a cache hit. I deemed this as more efficient. Also I do not think that a really large number of acquisition methods is that likely.

Or am I getting sth wrong here? Not opposed to changing it if you want to keep the filtering logic out of the hook.

src/common/POLFields/DetailsFields/FieldAcquisitionMethod.js
slaemmer: I tried `getFieldState() `, but it doesn't work: it returns undefined on the initial render, because `getFieldState` is a passive lookup `(state.fields[name]?.lastFieldState)`. It doesn't register anything itself, it just checks whether something else already has. In this component, the only other registrant is `FieldSelectionFinal`, which is created further down in the same render and hasn't mounted (registered) yet at the point where we read the value.

Practically, that means editing a line whose method was later deprecated could show the dropdown missing that value on first render, self-correcting only if something else happens to re-render the field.

`useField` avoids this because it's self-sufficient rather than relying on another component's registration: its first-render value comes from a synchronous register > read > unregister it does internally (in `useField's` `useState` lazy initializer) before returning anything, so it never depends on `FieldSelectionFinal` having gone first.
 
I also looked at reading `initialValues` off the form directly via `useForm().getState()`, which does sidestep the registration-order issue - but it trades a correctness dependency on something else in this file for an invisible one on `LayerPOLine`/`OrderTemplatesEditor` always fully remounting on reinitialize, which isn't guaranteed by contract anywhere. Given the double registration itself is cheap, I think it would be safer to leave `useField` as-is.

usavkov-epam: What about `useFormState`? :)

slaemmer: Ah well, yes. Changed:D

usavkov-epam (approved):

PR #1875, opened 2026-07-21
UIOR-1569: Correct refresh of prefix and suffix fields

charismatic-claire:
## Purpose

https://folio-org.atlassian.net/browse/UIOR-1569

This is a bugfix for the following problem: Prefix and suffix do not correctly refresh on selection in the create and edit of a Purchase Order.

## Approach

How did we solve this? Well, `FieldPrefix` and `FieldSuffix` contained broken validation logic, s.t. we could run into cases where we need to validate fields that aren't present. To fix that, we put the condition `shouldRevalidatePoNumber` in place, taking care of that cases. That makes sure we only run the validation when `poNumber` was mounted and is actually present, s.t. it can really be used for validation.

### DIFF
--- a/src/common/POFields/FieldPrefix.js
+++ b/src/common/POFields/FieldPrefix.js
@@ -11,6 +11,7 @@ import { PO_FORM_FIELDS } from '../constants';
 const FieldPrefix = ({
   isNonInteractive = false,
   prefixes,
+  shouldValidate,
   ...rest
 }) => {
   return (
@@ -19,7 +20,7 @@ const FieldPrefix = ({
       isNonInteractive={isNonInteractive}
       label={<FormattedMessage id="ui-orders.orderDetails.orderNumberPrefix" />}
       name={PO_FORM_FIELDS.poNumberPrefix}
-      validateFields={[PO_FORM_FIELDS.poNumber]}
+      validateFields={shouldValidate ? [PO_FORM_FIELDS.poNumber] : []}
       {...rest}
     />
   );
--- a/src/common/POFields/FieldPrefix.test.js
+++ b/src/common/POFields/FieldPrefix.test.js
@@ -2,7 +2,20 @@ import { Form } from 'react-final-form';
 
 import { render, screen } from '@folio/jest-config-stripes/testing-library/react';
 
+import { PO_FORM_FIELDS } from '../constants';
 import FieldPrefix from './FieldPrefix';
+
+const mockFieldSelectFinal = jest.fn(({ label }) => label);
+
+jest.mock('@folio/stripes-acq-components', () => {
+  const React = jest.requireActual('react');
+  const PropTypes = jest.requireActual('prop-types');
+
+  return {
+    FieldSelectFinal: (props) => mockFieldSelectFinal(props),
+    fieldSelectOptionsShape: PropTypes.arrayOf(PropTypes.shape({})),
+  };
+});
 
 const defaultProps = {
   prefixes: [],
@@ -21,9 +34,29 @@ const renderFieldPrefix = (props = {}) => render(
 );
 
 describe('FieldPrefix', () => {
-  it('should render \'prefix\' field', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should render the prefix field', () => {
     renderFieldPrefix();
 
     expect(screen.getByText('ui-orders.orderDetails.orderNumberPrefix')).toBeInTheDocument();
+  });
+
+  it('should not validate PO number when shouldValidate is false', () => {
+    renderFieldPrefix({ shouldValidate: false });
+
+    expect(mockFieldSelectFinal).toHaveBeenCalledWith(expect.objectContaining({
+      validateFields: [],
+    }));
+  });
+
+  it('should validate PO number when shouldValidate is true', () => {
+    renderFieldPrefix({ shouldValidate: true });
+
+    expect(mockFieldSelectFinal).toHaveBeenCalledWith(expect.objectContaining({
+      validateFields: [PO_FORM_FIELDS.poNumber],
+    }));
   });
 });
--- a/src/common/POFields/FieldSuffix.test.js
+++ b/src/common/POFields/FieldSuffix.test.js
@@ -2,7 +2,19 @@ import { Form } from 'react-final-form';
 
 import { render, screen } from '@folio/jest-config-stripes/testing-library/react';
 
+import { PO_FORM_FIELDS } from '../constants';
 import FieldSuffix from './FieldSuffix';
+
+const mockFieldSelectFinal = jest.fn(({ label }) => label);
+
+jest.mock('@folio/stripes-acq-components', () => {
+  const PropTypes = jest.requireActual('prop-types');
+
+  return {
+    FieldSelectFinal: (props) => mockFieldSelectFinal(props),
+    fieldSelectOptionsShape: PropTypes.arrayOf(PropTypes.shape({})),
+  };
+});
 
 const defaultProps = {
   suffixes: [],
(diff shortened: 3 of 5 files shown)


src/common/POFields/FieldPrefix.js
```diff
@@ -13,13 +14,21 @@ const FieldPrefix = ({
   prefixes,
   ...rest
 }) => {
+  const { getFieldState } = useForm();
+  const poNumberFieldState = getFieldState?.(PO_FORM_FIELDS.poNumber);
+  const shouldRevalidatePoNumber = Boolean(
+    poNumberFieldState?.value !== undefined &&
+    poNumberFieldState?.value !== '' &&
+    poNumberFieldState?.value !== null,
```
elsenhans: can be simplified to `const shouldRevalidatePoNumber = poNumberFieldState?.value;`

elsenhans:
Please add ticket to `CHANGELOG.md`

src/common/POFields/FieldPrefix.js
charismatic-claire: Good point, simplified.

src/common/POFields/FieldPrefix.test.js
```diff
@@ -1,9 +1,33 @@
...
+jest.mock('@folio/stripes-acq-components', () => {
+  const React = jest.requireActual('react');
+  const PropTypes = jest.requireActual('prop-types');
+
+  return {
+    FieldSelectFinal: ({ label, validateFields }) => (
+      <div data-testid="field-select" data-validate-fields={JSON.stringify(validateFields ?? [])}>
+        {label}
+      </div>
+    ),
```
elsenhans:
```suggestion
    FieldSelectFinal: (props) => mockFieldSelectFinal(props),
```

src/common/POFields/FieldPrefix.test.js
```diff
@@ -1,9 +1,33 @@
...
 
+jest.mock('react-final-form', () => {
+  const actual = jest.requireActual('react-final-form');
+
+  return {
+    ...actual,
+    useForm: jest.fn(),
+  };
+});
+
```
elsenhans:
```suggestion
const mockFieldSelectFinal = jest.fn(({ label }) => label);

```

src/common/POFields/FieldPrefix.test.js
```diff
@@ -21,9 +45,33 @@ const renderFieldPrefix = (props = {}) => render(
...
 
     expect(screen.getByText('ui-orders.orderDetails.orderNumberPrefix')).toBeInTheDocument();
   });
+
+  it('should not revalidate the PO number field when it is not mounted or empty', () => {
+    useForm.mockReturnValue({ getFieldState: () => undefined });
+
+    renderFieldPrefix();
+
+    expect(screen.getByTestId('field-select').dataset.validateFields).toBe('[]');
```
elsenhans: Please avoid the use of `data-testid` and `getByTestId`. Use DOM-based queries (e.g. `getByRole`, `getByLabelText`, `getByText`) instead or assert on the props passed to the mocked component.

```suggestion
expect(mockFieldSelectFinal).toHaveBeenCalledWith(expect.objectContaining({ validateFields: [] }));
```

src/common/POFields/FieldPrefix.test.js
```diff
@@ -21,9 +45,33 @@ const renderFieldPrefix = (props = {}) => render(
...
+  });
+
+  it('should revalidate the PO number field when it is mounted and has a value', () => {
+    useForm.mockReturnValue({
+      getFieldState: () => ({ name: PO_FORM_FIELDS.poNumber, value: '1001' }),
+    });
+
+    renderFieldPrefix();
+
+    expect(screen.getByTestId('field-select').dataset.validateFields).toBe(JSON.stringify([PO_FORM_FIELDS.poNumber]));
```
elsenhans:
```suggestion
    expect(mockFieldSelectFinal).toHaveBeenCalledWith(expect.objectContaining({
      validateFields: [PO_FORM_FIELDS.poNumber],
    }));
```

src/common/POFields/FieldSuffix.test.js
```diff
@@ -1,9 +1,33 @@
...
+  };
+});
+
+jest.mock('@folio/stripes-acq-components', () => {
+  const React = jest.requireActual('react');
+  const PropTypes = jest.requireActual('prop-types');
+
+  return {
+    FieldSelectFinal: ({ label, validateFields }) => (
+      <div data-testid="field-select" data-validate-fields={JSON.stringify(validateFields ?? [])}>
```
elsenhans: Same like for FieldPrefix.test.js:
> Please avoid the use of `data-testid` and `getByTestId`. Use DOM-based queries (e.g. `getByRole`, `getByLabelText`, `getByText`) instead or assert on the props passed to the mocked component.

src/common/POFields/FieldPrefix.test.js
```diff
@@ -21,9 +45,33 @@ const renderFieldPrefix = (props = {}) => render(
...
+
+  it('should render the prefix field', () => {
+    useForm.mockReturnValue({ getFieldState: () => undefined });
+
     renderFieldPrefix();
 
     expect(screen.getByText('ui-orders.orderDetails.orderNumberPrefix')).toBeInTheDocument();
   });
+
+  it('should not revalidate the PO number field when it is not mounted or empty', () => {
```
elsenhans: Empty string is never actually tested. The test is named "_should not revalidate ... when it is not mounted or empty_", but it only covers the "not mounted" case, never the case where the field is mounted with value: ''. 
Please consider adding a test for an empty value and/or renaming the test.

src/common/POFields/FieldSuffix.test.js
```diff
@@ -21,9 +45,33 @@ const renderFieldSuffix = (props = {}) => render(
...
+
+  it('should render the suffix field', () => {
+    useForm.mockReturnValue({ getFieldState: () => undefined });
+
     renderFieldSuffix();
 
     expect(screen.getByText('ui-orders.orderDetails.orderNumberSuffix')).toBeInTheDocument();
   });
+
+  it('should not revalidate the PO number field when it is not mounted or empty', () => {
```
elsenhans: Same here:
> Empty string is never actually tested. The test is named "_should not revalidate ... when it is not mounted or empty_", but it only covers the "not mounted" case, never the case where the field is mounted with value: ''.
Please consider adding a test for an empty value and/or renaming the test.

elsenhans (changes requested):

charismatic-claire:
@elsenhans Do you like the changes?

elsenhans:
> @elsenhans Do you like the changes?

Looking good. Thanks for the additional test case.
Some small points:
1. Seems like line 20 is unused:
`const React = jest.requireActual('react');`
Please remove it.

2. Fix eslint errors in `FieldPrefix.test.js` and `FieldSuffix.test.js`

NikitaSedyx (changes requested):
please address linter issues

additionally I didn't get how validation exclusion when po number is present solves your issue, I believe you just need to make component rerender and do not touch validation

charismatic-claire:
@elsenhans Thanks for finding this, I'll address the issue.
@NikitaSedyx I will address the linter issues. On your second question: I think the thrown exceptions due to the throwing validation (because it cannot grab the value to validate) IS THE CAUSE of the issue of the field not refreshing properly. See the video attached. Otherwise I wouldn't expect that an input field update requires a manual refresh. The reactive nature of, well... React, should take care of that. I think.

[Screencast_20260723_123408.webm](https://github.com/user-attachments/assets/c3073a52-757e-4b0b-84cd-57f6d51281e9)

charismatic-claire:
@elsenhans I adressed your comments.

@NikitaSedyx I fixed all the linter issues including two unrelated ones. I added an extra commit for it to indicate the fix doesn't belong to the ticket itself.

src/components/POLine/POLineAgreementLines/POLineAgreementLinesContainer.js
```diff
@@ -49,7 +49,7 @@ const POLineAgreementLinesContainer = ({ lineId, label, mutator }) => {
     [fetchAgeementLines],
   );
 
-  useEffect(refreshList, [lineId]);
+  useEffect(refreshList, [lineId, refreshList]);
```
charismatic-claire: @NikitaSedyx out of scope linter fix in extra commit. If that's considered bad practice, we can remove it.

src/components/PurchaseOrder/POForm.js
```diff
@@ -120,7 +120,7 @@ const POForm = ({
         ...(orderTemplate?.hiddenFields || {}),
       }));
     }
-  }, []);
+  }, [initialValues.template, parentResources?.orderTemplates?.records]);
```
charismatic-claire: @NikitaSedyx out of scope linter fix in extra commit. If that's considered bad practice, we can remove it.

NikitaSedyx: I would suggest to not fix issues not connected to your changes

NikitaSedyx:
> I think the thrown exceptions due to the throwing validation (because it cannot grab the value to validate) IS THE CAUSE of the issue of the field not refreshing properly. See the video attached. Otherwise I wouldn't expect that an input field update requires a manual refresh. The reactive nature of, well... React, should take care of that. I think.

hm, in some extent it makes sense, but I would like to play a bit - it's pretty old code so I do not remember all details

charismatic-claire:
@NikitaSedyx did you take a look and could reproduce the bug and verify the fix?

src/components/PurchaseOrder/POForm.js
charismatic-claire: Alrighty

NikitaSedyx (approved):

elsenhans (approved):
LG

